PR #1, opened 2020-12-15
Add main classes for query

evanh:
Add the main classes for Query and the supporting Expressions, as well as
minimal/stub functions for validation and translation. The idea behind
validation is that it should either run or throw an Exception. The translation
function is very minimal and needs to be cleaned up.


mypy.ini
```diff
@@ -10,7 +10,7 @@ disallow_any_generics = True
 disallow_incomplete_defs = True
 disallow_subclassing_any = True
 ; disallow_untyped_calls = True
-disallow_untyped_decorators = True
+disallow_untyped_decorators = False
```
fpacifici, line 13: Why this ?

snuba_sdk/__init__.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class Expression(ABC):
+    def __post_init__(self) -> None:
+        self.validate()
+
+    @abstractmethod
+    def validate(self) -> None:
+        raise NotImplementedError
```
fpacifici, line 10: Are you sure an expression will be able to validate itself in isolation without context? 
Defining the method as part of the query structure means that either you provide context in the object to validate itself (which seems unlikely since they are immutable) or that validation is executed by each node independently without context.
This may or may not be possible depending on what you plan to validate. If you try to validate that a given column is in the logical schema, you need more context when calling validate on the column object.

snuba_sdk/expressions.py
```diff
@@ -0,0 +1,109 @@
...
+    pass
+
+
+@dataclass(frozen=True)
+class Limit(Expression):
+    limit: int
+
+    def validate(self) -> None:
+        if self.limit < 0 or self.limit > 10000:
+            raise InvalidExpression(f"limit {self.limit} is invalid")
```
fpacifici, line 23: why not putting these type of structural validation in the _post_init method? A limit object with limit lower than 0 should never be instantiated.

snuba_sdk/query.py
```diff
@@ -0,0 +1,99 @@
...
+        for exps in clauses:
+            if exps is not None:
+                for exp in exps:
+                    exp.validate()
+
+        self.limit.validate() if self.limit is not None else None
+        self.offset.validate() if self.offset is not None else None
+        self.granularity.validate() if self.granularity is not None else None
+
+    def translate(self) -> str:
```
fpacifici, line 78: If you were talking about this when you were considering using a visitor, strong +1 on that. 
That would allow you separate the translation from the query structure itself, which means you can swap different translators (imagine a pretty print one with a simpler one) without bloating the structure of your query and all the expressions. This can blow up quickly.
Another use case could be allowing you to have a specific serializer for logs that does not serialize literals so identical queries with different project and time range are serialized in the same way.

fpacifici:
Some early comments, hope they are helpful.

mypy.ini
evanh: To avoid the `pytest.mark` warning. It's the only decorator I have.

snuba_sdk/__init__.py
evanh: My plan is to let the expressions validate as much as possible independently, and then have some contextual validators at the query level. For example, a function can validate itself and then ensure its arguments are valid, but the query needs to assert that top level functions in the SELECT clause have aliases.

snuba_sdk/expressions.py
evanh: The __post_init__ calls `self.validate` so that should already be true.

snuba_sdk/query.py
evanh: Yeah OK. I'll have to add that in. Even with just these two cases (validate and translate) it feels cumbersome.

snuba_sdk/conditions.py
```diff
@@ -0,0 +1,55 @@
...
+                f"invalid condition: LHS of a condition must be a Column or Function, not {type(self.lhs)}"
+            )
+        if not isinstance(self.rhs, (Column, Function, *Scalar)):
+            raise InvalidExpression(
+                f"invalid condition: RHS of a condition must be a Column, Function or Scalar not {type(self.rhs)}"
+            )
+        if not isinstance(self.op, Op):
+            raise InvalidExpression(
+                "invalid condition: operator of a condition must be an Op"
+            )
```
fpacifici, line 47: Are you validating types since sentry does not use mypy ?

evanh: Essentially, although I can't guarantee Sentry will be the only user of this package, so I think it's smart to do type checking on all the external interfaces.

PR #3, opened 2021-01-06
fix(query) Add more validation and tests for Query

evanh:
Add some contextual validation to the Query and some tests for different
error states.

### DIFF
--- a/snuba_sdk/query.py
+++ b/snuba_sdk/query.py
@@ -19,7 +19,8 @@ class InvalidQuery(Exception):
 
 
 def list_type(vals: List[Any], type_classes: Sequence[Any]) -> bool:
-    return not isinstance(vals, list) or not all(
+    print("1", isinstance(vals, list))
+    return isinstance(vals, list) and all(
         isinstance(v, tuple(type_classes)) for v in vals
     )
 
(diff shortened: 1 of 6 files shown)


snuba_sdk/query.py
```diff
@@ -83,18 +84,56 @@ def set_granularity(self, granularity: int) -> Query:
...
+            for group_exp in non_aggregates:
+                if group_exp not in self.groupby:
+                    raise InvalidQuery(
+                        f"{group_exp.translate()} missing from the groupby"
+                    )
+
+        # TODO - It's not clear if this is worth doing. Each of these components was validated
+        # when they were created, so do we need to validate again? It is possible for someone
+        # to override a component with something invalid, but if there is only so much
+        # idiot proofing that can be done with Python.
```
fpacifici, line 123: If your dataclass is frozen it is immutable. If the collections inside the class are not MutableSequence, MutableMapping etc. they are immutable, if you change the content in place you will at least have a type error.
So you should not need to validate structures that are immutable. Like Match.

Now, when to validate the entire query is a different and more interesting problem. I would rather we decide a validation policy that is comprehensive and we are convinced that is needed and stick to that. This would make it much easier to have a good confidence on the validity status of the query when dealing with that.

I think deciding doing a deep validation when we validate the query class is a good idea. Redoing it when formatting and translating seems a good option to me to catch improper usages of the api.
IF we guarantee these two are done, then everything else in between is a plus and we would not have to worry when maintaining the api.

fpacifici (commented):
I'd rather we did not merge without code reviews.

PR #4, opened 2021-01-07
feat(visitor) Add a visitor pattern to the Query

evanh:
Add a new Visitor class generic that can be extended for different ways to
traverse a query. Also add a validation and translation version of the visitor
and use that instead of the `validate` and `translate` functions.


snuba_sdk/expressions.py
```diff
@@ -1,16 +1,52 @@
...
-from snuba_sdk import Expression
 from snuba_sdk.clickhouse import is_aggregation_function
 
 
-class InvalidExpression(Exception):
-    pass
+TVisited = TypeVar("TVisited")
+
+
+class Visitor(ABC, Generic[TVisited]):
```
evanh, line 16: The downside of this is that since the base Expression class wants to validate and translate itself, all of this code has to be in the same file to avoid circular imports. I think future visitors should be able to be in a separate file though.

fpacifici: The other alternative is to mimic pattern matching instead of implementing a visitor, just as we do for query classes in snuba: 
https://github.com/getsentry/snuba/blob/master/snuba/query/data_source/visitor.py#L11-L61

This is basically copying over what the visitor does. In language with pattern matching support generally that would give us the same type safety as a visitor (type checking would fail if we add a new type of query, expression here). But there is a way to mimic that in python through mypy exhaustive checks: https://hakibenita.com/python-mypy-exhaustive-checking#type-narrowing-in-mypy
I was not aware of this when I added the data source visitor mimicking pattern matching.

snuba_sdk/expressions.py
```diff
@@ -130,44 +130,163 @@ class Function(Expression):
...
+
+class InvalidExpression(Exception):
+    pass
+
+
+class InvalidEntity(Exception):
+    pass
+
+
+class Validation(Visitor[bool]):
```
fpacifici, line 179: IF you are rising exceptions when the validation fails instead of returning false you should make this:
`class Validation(Visitor[None])`

otherwise you have three expected return conditions: True, False, exception, while you need only two.

snuba_sdk/expressions.py
```diff
@@ -130,44 +130,163 @@ class Function(Expression):
...
+            )
+        if not isinstance(cond.op, Op):
+            raise InvalidExpression(
+                "invalid condition: operator of a condition must be an Op"
+            )
+
+        return True
+
+
+class Translation(Visitor[str]):
```
fpacifici, line 263: Did you keep the visitor in the same module for a specific reason? While the Visitor abstract class needs to stay in the same module as the visited class with this implementation, the implementations of the visitor do not have the same constraint and can be anywhere.

snuba_sdk/query.py
```diff
@@ -3,28 +3,33 @@
...
         isinstance(v, tuple(type_classes)) for v in vals
     )
 
 
+TRANSLATOR = Translation()
+VALIDATOR = Validation()
+
+
 @dataclass(frozen=True)
 class Query:
```
fpacifici, line 34: I cannot remember  if there was a reason for not making Query an expression. Or not having a common class on top of Query and expression (like Visitable) that would allow to use the same approach to navigate through the tree for Queries and expressions. That would mean learning one way instead of two.
In Snuba this was for historic reasons, Query was too tangled with the dataset logic at the time ew introduced the AST. The second reason, that would be present here as well, is that turning a query into an expression is not enough, as we do not want to have an argument of a function call to be a Query, so there has to be an intermediate class.

snuba_sdk/expressions.py
evanh: Not quite, because the base Expression class uses both the Validation and Translation visitors to automatically validate/translate. I could probably rewrite this to move the translation one out of this file, but the validation one would need to stay.

snuba_sdk/query.py
evanh: I have a comment in the validation function I think, but the main reason the Query isn't an Expression is validation. An Expression is assumed to be always valid, since I check its validity in the __post_init__. However a Query can exist in an invalid state, if for example you initialize it and then add the select columns later. While waiting for the select columns the query is in an invalid state. Once I add subquery support  the idea of making a Query visitable will be needed.

snuba_sdk/expressions.py
fpacifici: Right... meh.
Let's take a step back. The reason we started talking about visitors (or pattern matching if you prefer that alternative) is to decouple a feature that needs to traverse the tree from the core node classes themselves. This allows you not to squeeze all those responsibilities into the core classes themselves, which would soon explode in complexity.
Now it seems this works well for translators, as you can have multiple translation mechanisms (think about pretty print vs simple compact format vs formatting for tracing that removes literals). And the fact you can move that class out of this module seems a good sign.

If validation is inherently coupled with the core classes because you call validation in post_init, decoupling it is not that relevant. The question is another one though: how much logic do you expect to add in the post_init validation ? So far it seems simple enough that extracting from the core classes or not is not a major change. But I remember we talked about contextual validation. Would that still be tightly coupled with the core classes or would it be separate ?

evanh: We can keep contextual validation at the Query level for now, or maybe as separate visitors. So I think the validation on the individual expressions should remain fairly static. I'll remove the Validation visitor, that should clean this up quite a bit.

fpacifici: Make sense to me, though I think we may need to decide sooner than later what kind of contextual validation we want to implement and whether we allow the client to plug their own validation logic (in could be interesting to let the client define dataset specific validation, but I did not put a lot of thinking in it).

snuba_sdk/expressions.py
evanh: I ended up using the pattern matching idea, as well as moving basic validation back into the Expression class. Things are now split into modules, which is good because the Condition and Entity logic will get a lot more complicated in the future. Not a perfect visitor pattern but what are you gonna do.

fpacifici (approved):
Good to me.
Two remarks:
- please see my comment on the exhaustive check. It should be possible to make the if block in the visitor exhaustive so that we preserve the advantage of the visitor structure, which is to notice at type checking time if we add a subclass of expression and forget about the visitor.
- Since we are about to start plugging this into Sentry code, I think it is time to make code review a requirement for merging.

snuba_sdk/visitors.py
```diff
@@ -0,0 +1,128 @@
...
+        elif isinstance(node, Condition):
+            return self._visit_condition(node)
+        elif isinstance(node, Limit):
+            return self._visit_int_literal(node.limit)
+        elif isinstance(node, Offset):
+            return self._visit_int_literal(node.offset)
+        elif isinstance(node, Granularity):
+            return self._visit_int_literal(node.granularity)
+
+        raise NotImplementedError
```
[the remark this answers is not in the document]
fpacifici: This explains all the details about the exhaustive checks https://hakibenita.com/python-mypy-exhaustive-checking#type-narrowing-in-mypy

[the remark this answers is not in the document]
evanh: Oh whoops yes I read that article and changed this but forgot to push the change. I'll update.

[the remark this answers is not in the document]
evanh: I wasn't able to use the `assert_never` function because this function takes an Expression, and mypy isn't smart enough to know that I have checked for all subclasses of Expression, which is itself an ABC. I don't really want to have a giant `Union[Column, Function...]` in the argument types either.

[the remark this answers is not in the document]
fpacifici: ok, I give up, this is python telling us: "I don't have pattern matching. Stop trying to use it."

PR #5, opened 2021-01-08
feat(open source) Add infrastructure for proper builds/releases

evanh:
Add Craft and other workflows to allow proper releases of the project.


.craft.yml
```diff
@@ -0,0 +1,12 @@
+minVersion: "0.15.0"
+github:
+  owner: getsentry
+  repo: snuba-sdk
+releaseBranchPrefix: releases
```
BYK, line 5: I'd drop this and use the default (which is `release` and that's what you use below)

.craft.yml
```diff
@@ -0,0 +1,12 @@
...
+changelog: CHANGES
+changelogPolicy: auto
+artifactProvider:
+  name: github
+targets:
+  - name: github
+  - name: pypi
```
BYK, line 12: You may want to use `requiredNames` to make sure you have _some_ artifacts, like we do over at Sentry: https://github.com/getsentry/sentry/blob/1a539eaf3fed64a7085b80166be90bc6d20b1f14/.craft.yml#L38-L39

.github/workflows/build.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - run: |
+          python -m pip install --upgrade pip
```
BYK, line 22: This may not be a great idea. Why always use the latest `pip`, instead of pinning a version?

.github/workflows/build.yml
```diff
@@ -0,0 +1,28 @@
...
+          pip install virtualenv wheel
```
BYK, line 23: Do you really need `virtualenv`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,28 @@
...
+          python setup.py sdist bdist_wheel
```
BYK, line 24:
```suggestion
          python setup.py bdist_wheel
```

I don't think we need an `sdist`, do we?

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    name: 'Release a new version'
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GH_RELEASE_PAT }}
+          fetch-depth: 0
```
BYK, line 21: You should be able to drop this one.

BYK (changes requested):
Requesting changes for the `releaseBranchPrefix` fix as without that fix, this whole thing wouldn't work.

.github/workflows/build.yml
evanh: I copied this from the GHA Python template. I have no strong feelings about it, just one less package I have to update periodically.

.github/workflows/build.yml
evanh: Ah no that was from an older version.

.github/workflows/build.yml
evanh: I guess not.

.github/workflows/build.yml
BYK: I'm fairly sure this will break your builds unexpectedly some time in the future but it should be rare :)

BYK (approved):
Looks good. Hope this works properly 🤞🏻

PR #10, opened 2021-01-13
feat(query) Create Query visitor, add missing clauses

evanh:
- Add the last few missing clauses: having, orderby, limitby, totals.
- I realized that the query needed to be translated into a JSON blob, since
that's what Snuba ultimately expects. That prompted the realization that it
was time to create a QueryVisitor class.
- Add a pretty print and snuba query visitor.

### DIFF
--- a/snuba_sdk/expressions.py
+++ b/snuba_sdk/expressions.py
@@ -5,7 +5,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import date, datetime
 from enum import Enum
-from typing import List, Optional, Set, Union
+from typing import Any, List, Optional, Sequence, Set, Union
 
 from snuba_sdk.snuba import check_array_type, is_aggregation_function
 
--- a/snuba_sdk/expressions.py
+++ b/snuba_sdk/expressions.py
@@ -136,7 +136,14 @@ class Function(Expression):
     alias: Optional[str] = None
 
     def is_aggregate(self) -> bool:
-        return is_aggregation_function(self.function)
+        if is_aggregation_function(self.function):
+            return True
+
+        for param in self.parameters:
+            if isinstance(param, Function) and param.is_aggregate():
+                return True
+
+        return False
 
     def validate(self) -> None:
         if not isinstance(self.function, str):
--- a/snuba_sdk/expressions.py
+++ b/snuba_sdk/expressions.py
@@ -4,7 +4,8 @@ import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import date, datetime
-from typing import Any, List, Optional, Sequence, Set, Union
+from enum import Enum
+from typing import List, Optional, Set, Union
 
 from snuba_sdk.snuba import check_array_type, is_aggregation_function
 
@@ -106,6 +107,15 @@ class Granularity(Expression):
 
 
 @dataclass(frozen=True)
+class Totals(Expression):
+    totals: bool = False
+
+    def validate(self) -> None:
+        if not isinstance(self.totals, bool):
+            raise InvalidExpression("totals must be a boolean")
+
+
+@dataclass(frozen=True)
 class Column(Expression):
     name: str
 
@@ -156,3 +166,41 @@ class Function(Expression):
                 raise InvalidExpression(
                     f"parameter '{param}' of function {self.function} is an invalid type"
                 )
+
+    def __eq__(self, other: object) -> bool:
+        # Don't use the alias to compare equality
+        if not isinstance(other, Function):
+            return False
+
+        return self.function == other.function and self.parameters == other.parameters
+
+
+class Direction(Enum):
+    ASC = "ASC"
+    DESC = "DESC"
+
+
+@dataclass(frozen=True)
+class OrderBy(Expression):
+    exp: Union[Column, Function]
+    direction: Direction
+
+    def validate(self) -> None:
+        if not isinstance(self.exp, (Column, Function)):
+            raise InvalidExpression("OrderBy expression must be a Column or Function")
+        if not isinstance(self.direction, Direction):
+            raise InvalidExpression("OrderBy direction must be a Direction")
+
+
+@dataclass(frozen=True)
+class LimitBy(Expression):
+    column: Column
+    count: int
+
+    def validate(self) -> None:
+        if not isinstance(self.column, Column):
+            raise InvalidExpression("LimitBy can only be used on a Column")
+        if not isinstance(self.count, int) or self.count <= 0 or self.count > 10000:
+            raise InvalidExpression(
+                "LimitBy count must be a positive integer (max 10,000)"
+            )
--- a/snuba_sdk/query_visitors.py
+++ b/snuba_sdk/query_visitors.py
@@ -5,7 +5,7 @@ from abc import ABC, abstractmethod
 from typing import (
     Generic,
     List,
-    MutableMapping,
+    Mapping,
     Optional,
     TypeVar,
     TYPE_CHECKING,
@@ -39,29 +39,15 @@ QVisited = TypeVar("QVisited")
 
 class QueryVisitor(ABC, Generic[QVisited]):
     def visit(self, query: Query) -> QVisited:
-        returns = {
-            "dataset": self._visit_dataset(query.dataset),
-            "match": self._visit_match(query.match),
-            "select": self._visit_select(query.select),
-            "groupby": self._visit_groupby(query.groupby),
-            "where": self._visit_where(query.where),
-            "having": self._visit_having(query.having),
-            "orderby": self._visit_orderby(query.orderby),
-            "limitby": self._visit_limitby(query.limitby),
-            "limit": self._visit_limit(query.limit),
-            "offset": self._visit_offset(query.offset),
-            "granularity": self._visit_granularity(query.granularity),
-            "totals": self._visit_totals(query.totals),
-        }
+        fields = query.get_fields()
+        returns = {}
+        for field in fields:
+            returns[field] = getattr(self, f"_visit_{field}")(getattr(query, field))
 
         return self._combine(query, returns)
 
     @abstractmethod
-    def _combine(
-        self,
-        query: Query,
-        returns: MutableMapping[str, QVisited],
-    ) -> QVisited:
+    def _combine(self, query: Query, returns: Mapping[str, QVisited]) -> QVisited:
         raise NotImplementedError
 
     @abstractmethod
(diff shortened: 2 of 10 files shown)


snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+
+class InvalidQuery(Exception):
+    pass
+
+
+QVisited = TypeVar("QVisited")
+
+
+class QueryVisitor(ABC, Generic[QVisited]):
+    def visit(self, query: Query) -> QVisited:
```
evanh, line 41: I spent a bunch of time trying to figure out if I could get mypy to show an error if the fields of the query were modified. I'm still open to suggestions about how to do that.

fpacifici: >  if I could get mypy to show an error if the fields of the query were modified

This is not possible. mypy does not act at runtime, thus there is no way for it to assert that a call to a method returns a different object with respect to the one you passed. It can only check types statically. 
The only way I know of to avoid that kind of issue (passing an object to a method and ensure it is not changed) is by making the object immutable and ensure mypy is used so any attempt of mutation is caught by the type checker. 
A less compelling alternatives is passing copies of the objects you want to ensure are not changed and discard the copy afterwards.

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+    @abstractmethod
+    def _combine(
+        self,
+        query: Query,
+        returns: MutableMapping[str, QVisited],
+    ) -> QVisited:
+        raise NotImplementedError
+
+    @abstractmethod
+    def _visit_dataset(self, dataset: str) -> QVisited:
```
fpacifici, line 68: I think assuming the type returned for each clause before calling combine will be the same as the type returned after combine may be a strong assumption. Though you can definitively change this later without too many issues.

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+            "groupby": self._visit_groupby(query.groupby),
+            "where": self._visit_where(query.where),
+            "having": self._visit_having(query.having),
+            "orderby": self._visit_orderby(query.orderby),
+            "limitby": self._visit_limitby(query.limitby),
+            "limit": self._visit_limit(query.limit),
+            "offset": self._visit_offset(query.offset),
+            "granularity": self._visit_granularity(query.granularity),
+            "totals": self._visit_totals(query.totals),
+        }
```
fpacifici, line 55: The clauses are statically defined, they will always be the same and always all visited. It seems this structure should be a NamedTuple instead of a dict.

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def _visit_groupby(
+        self, groupby: Optional[List[Union[Column, Function]]]
+    ) -> QVisited:
+        raise NotImplementedError
+
+    @abstractmethod
+    def _visit_where(self, where: Optional[List[Condition]]) -> QVisited:
```
fpacifici, line 88: Why a list and not a Sequence? (list is mutable, sequence is not).

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+            "limitby",
+            "limit",
+            "offset",
+            "granularity",
+            "totals",
+        ]
+        separator = "\n" if self.pretty else " "
+        formatted = separator.join([returns[c] for c in clauses if returns[c]])
+        if self.pretty:
+            formatted = f"-- DATASET: {returns['dataset']}\n{formatted}"
```
fpacifici, line 142: Do we actually parse comments in SnQL ?  It would be nice to keep the query executable when we pretty print it.

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+        return ""
+
+    def _visit_where(self, where: Optional[List[Condition]]) -> str:
+        if where is not None:
+            return f"WHERE {' AND '.join(self.translator.visit(w) for w in where)}"
+        return ""
+
+    def _visit_having(self, having: Optional[List[Condition]]) -> str:
+        if having is not None:
+            return f"HAVING {', '.join(self.translator.visit(h) for h in having)}"
```
fpacifici, line 169: Don't you have to join the HAVING conditions with an AND ?

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+    def _visit_totals(self, granularity: Optional[Totals]) -> QVisited:
+        raise NotImplementedError
+
+
+class Printer(QueryVisitor[str]):
+    def __init__(self, pretty: bool = False) -> None:
+        self.translator = Translation()
+        self.pretty = pretty
+
+    def _combine(self, query: Query, returns: MutableMapping[str, str]) -> str:
```
fpacifici, line 125: Why a MutableMapping ? Can't it be immutable ?

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+            raise InvalidQuery("query must have at least one column in select")
+
+        # - Orderby must be a field in select
+        if query.orderby is not None:
+            for o in query.orderby:
+                found = False
+                for s in query.select:
+                    if s == o.exp:
+                        found = True
+                        break
```
fpacifici, line 229: Why do we enforce that the order by is in the select ? A query that sorts according to a key and then return other fields does not seem necessarily invalid to me.

Also I don't think this works. 
If I have `SELECT a, b ,c`
and `ORDER BY a, x, c`

If you want to ensure all columns in the order by are also in the select, this check will pass while it should not.

snuba_sdk/query_visitors.py
```diff
@@ -0,0 +1,321 @@
...
+                )
+
+        # Top level functions in the select clause must have an alias
+        non_aggregates = []
+        has_aggregates = False
+        for exp in query.select:
+            if isinstance(exp, Function) and not exp.alias:
+                raise InvalidQuery(f"{exp} must have an alias in the select")
+
+            if not isinstance(exp, Function) or not exp.is_aggregate():
```
fpacifici, line 256: how does this identify aggregates inside a function?
`plus(count(event_id), count(trace_id))`

fpacifici (commented):
I would consider keeping unrelated changes in separate PRs going on (adding fields to the query and adding the query visitor/pretty print). It makes it hard to tell what changes depends on what.

snuba_sdk/query_visitors.py
evanh: We don't, although it would be trivial to add.

snuba_sdk/expressions.py
```diff
@@ -116,3 +126,41 @@ def validate(self) -> None:
                 raise InvalidExpression(
                     f"parameter '{param}' of function {self.function} is an invalid type"
                 )
+
+    def __eq__(self, other: object) -> bool:
+        # Don't use the alias to compare equality
```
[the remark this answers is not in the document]
evanh: I couldn't think of a scenario where we would want to include the alias in the equality. Since the semantic equality seems to be more used, I made that the default. If a strong equality is needed I can add it back.

snuba_sdk/query_visitors.py
evanh: I wasn't talking about at runtime, I meant a mypy compile error. If you are editing the code and you add a field to the query, you have to add it to a number of other functions and I wanted mypy to warn you of that.

snuba_sdk/query_visitors.py
evanh: This grew out of the current use cases. Although I did debate having it be different for the Translator, so it would return a dict instead of a JSON string. At the end of the day I thought this would be simpler.

snuba_sdk/query_visitors.py
evanh: These match the types in the Query object, which specifies a list. Passing a tuple to this function would be incorrect.

snuba_sdk/query_visitors.py
evanh: That's my mistake I thought that would be invalid but you're right clickhouse allows ordering by non-selected fields. I'll remove this check.

snuba_sdk/query_visitors.py
evanh: Changing this to a NamedTuple requires creating a Generic named tuple class and then creating a new version for every return type. I think this is simpler.

fpacifici: A generic class can be concrete. Your named tuple does not need to be abstract:

In theory you should be able to do this.
```
TField = TypeVar("TField")
class PreCombineQuery(Generic[TField], NamedTuple): # Not sure which order the generic and named tuple should be in the declaration. And you can do the same with a dataclass.
   dataset: TField
   match: TField
   ..... 
```

In practice both named tuples and dataclasses need some more help to get the type right:
```
class PreCombineQuery(Generic[TField], NamedTuple):
       dataset: TField

       get_dataset(self) -> TField
```

snuba_sdk/query_visitors.py
fpacifici: Ah so you were not talking about visit methods mutating the content of the argument they receive but the structure of the query changing by adding fields. Sorry my bad.
Then I am not sure about the problem. If you add a new field to the query, for as long as you add it to the abstract visitor and the visit method of the abstract calls it, all subclasses will need to implement that for mypy to be ok. (you may have to drop the idea of  dynamically building the method name you call since mypy cannot check that).
If instead the issue is in ensuring that `combine` methods take the new field into account, then there is no way neither with this architecture nor by adding everything in the query class. You can make it harder to miss by exploding the arguments to the combine method. Instead of passing an opaque dict (or event a namedtuple or dataclass), pass each individual argument. This forces you to touch that method in every visitor implementation and at least think about it.

snuba_sdk/query_visitors.py
evanh: So I did try this, and while I could create the Generic class, I couldn't go further. Here's one example below with the mypy errors, although I tried a number of different ways to see if I could fix the errors.

```
from typing import TypeVar, Generic, NamedTuple


TField = TypeVar("TField")


class Parts(NamedTuple, Generic[TField]):
    part1: TField
    part2: TField

# "Parts" expects no type arguments, but 1 given
# Generic tuple types not supported
# Class cannot subclass value of type "Any"
class StrParts(Parts[str]):
    pass


p = StrParts("1", "2")
```

snuba_sdk/query_visitors.py
evanh: It's not a problem, I just thought it would be nice if when you add a new field to the query, mypy will warn you about all the other places you need to add the field.

snuba_sdk/query_visitors.py
fpacifici: ok

snuba_sdk/query_visitors.py
evanh: @fpacifici I know we talked about a NamedTuple for this, but now that it's driven by the query itself I don't think we need an intermediate dataclass.

fpacifici (approved):

snuba_sdk/query_visitors.py
fpacifici: ok

PR #11, opened 2021-01-15
feat(arrays/tuples) Allow arrays and tuples as scalar values

evanh:
Allow passing arrays and tuples of scalar values as arguments. Also implement
a function that mimics the array typing rules that Clickhouse has to avoid
sending invalid array values.

### DIFF
--- a/snuba_sdk/clickhouse.py
+++ b/snuba_sdk/clickhouse.py
@@ -1,3 +1,5 @@
+from typing import Any, List, Optional
+
 # This is supposed to enumerate the functions snuba supports (with their
 # validator) so we can keep control of the clickhouse functions snuba
 # exposes.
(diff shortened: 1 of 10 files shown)


snuba_sdk/clickhouse.py
```diff
@@ -102,3 +104,49 @@
...
+    """
+    Check if a list follows the Clickhouse array typing rules.
+    - An array must contain all the same data type, or NULL
+    - An array can nest arrays, but those arrays must all hold the same data type
+    """
+
+    def find_base(value: Any) -> Optional[str]:
+        if value is None:
+            return None
+        elif isinstance(value, (int, float)):
```
fpacifici, line 119: you should be able to use `isinstance(value, numbers.Number)`

snuba_sdk/expressions.py
```diff
@@ -23,9 +23,45 @@ def validate(self) -> None:
...
+        if len(value_str) > 10:
+            value_str = f"{value_str[:10]}...]"
+
+        super().__init__(
+            f"invalid array {value_str}: arrays must have the same data type or None, perhaps use a tuple instead"
+        )
+
+
+def is_scalar(value: Any) -> bool:
+    if isinstance(value, tuple(Scalar)) or isinstance(value, tuple):
```
fpacifici, line 55: Why does `isinstance(value, tuple)`  make the method return true? a Tuple is not a scalar.

fpacifici (approved):
A couple of remarks in line

snuba_sdk/expressions.py
evanh: It is now, that's what this PR adds. list and tuple are now valid scalar types, as long as they themselves contain scalars. I will add a check for that, since I have it for the array.

PR #12, opened 2021-01-18
feat(json) Translate JSON snuba to SnQL

evanh:
Depends on #10 (also 75% of the line count is tests)

Add a function that takes a JSON snuba query and an entity, and converts it into
a SnQL SDK object that can then be translated to pure SnQL. This will be used to
help with the migration in Sentry and can be removed at a later date.


snuba_sdk/utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+        return value
+
+    dataset = body.get("dataset", "")
+    query = Query(dataset, Entity(entity))
+
+    selected_columns = list(map(to_exp, body.get("selected_columns", [])))
+    for a in body.get("aggregations", []):
+        if a[0].endswith(")") and not a[1]:
+            selected_columns.append(Function(a[0].strip("()"), [], a[2]))
```
fpacifici, line 56: I believe the visibility team is still using infix expressions in aggregations. Mind adding a check for such expressions (checking for the presence of parentheses after you removed these) and fail explicitly?

snuba_sdk/utils.py
```diff
@@ -0,0 +1,138 @@
...
+        limit, name = limitby
+        query = query.set_limitby(LimitBy(Column(name), int(limit)))
+
+    extras = ("limit", "offset", "granularity", "totals")
+    for extra in extras:
+        if body.get(extra) is not None:
+            query = getattr(query, f"set_{extra}")(body.get(extra))
+
+    # sample = legacy.get("sample")
+    # sample_clause = f"SAMPLE {sample}" if sample else ""
```
fpacifici, line 137: Why are these commented ?

fpacifici (commented):
Is the sampling clause intentionally missing ?

snuba_sdk/utils.py
evanh: I had a section here for things that we don't support yet. I'll remove for now.

snuba_sdk/utils.py
evanh: There's no need for an explicit exception, this will throw an InvalidExpression exception since "(" are not allowed in function names. However should I try to unravel the function? I'm just worried that if this function doesn't translate any JSON snuba query to SnQL, then the migration to SnQL will stall while we wait for the product teams to start using the SDK properly.

snuba_sdk/utils.py
evanh: I did intentionally leave out the SAMPLE clause in order to avoid feature creep. The point of this was to add the bare minimum to support the features used by the sessions queries. Since those don't use sample, I didn't add it.

snuba_sdk/utils.py
fpacifici: > There's no need for an explicit exception, this will throw an InvalidExpression exception since "(" are not allowed in function names.

That is what I meant by adding an explicit expression. Explain in the exception that the behavior is not supported. Otherwise people will keep trying to port those queries and complain when they get this cryptic InvalidExpression message.

> However should I try to unravel the function
If you are talking about parsing the function, I would discourage that as you would have to port the parser and its grammar here. But that's up to you.

fpacifici (approved):
I would still raise a clear error for the cases when the client passes an unsupported aggregation instead of relying on InvalidExpression. This would prevent people from thinking this is a bug.

PR #14, opened 2021-01-25
fix(legacy) Support function order by

evanh:
Correctly parse functions in the orderby from legacy Snuba queries.

### DIFF
--- a/snuba_sdk/legacy.py
+++ b/snuba_sdk/legacy.py
@@ -120,8 +120,12 @@ def json_to_snql(body: Mapping[str, Any], entity: str) -> Query:
 
         order_bys = []
         for o in order_by:
-            direction = Direction.DESC if o.startswith("-") else Direction.ASC
-            order_bys.append(OrderBy(Column(o.lstrip("-")), direction))
+            direction = Direction.ASC
+            if isinstance(o, str) and o.startswith("-"):
+                direction = Direction.DESC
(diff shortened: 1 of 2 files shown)


snuba_sdk/legacy.py
```diff
@@ -120,8 +120,12 @@ def to_scalar(value: Any) -> Any:
 
         order_bys = []
         for o in order_by:
-            direction = Direction.DESC if o.startswith("-") else Direction.ASC
-            order_bys.append(OrderBy(Column(o.lstrip("-")), direction))
+            direction = Direction.ASC
+            if isinstance(o, str) and o.startswith("-"):
```
lynnagara, line 124: Are we not already checking the type before this point? If not then shouldn't we throw an error here rather than silently ignore the invalid order by term?

evanh: > Are we not already checking the type before this point?

This is taking a query generated in Sentry, so we can't be sure that it has been properly validated.

> shouldn't we throw an error here

A list type here is valid, that's what this changes. The `to_exp` function will correctly convert lists and raw strings and throw an exception otherwise.

lynnagara (approved):

snuba_sdk/legacy.py
lynnagara: Ok

PR #15, opened 2021-01-25
fix(gha) Actions are not actually running tests

evanh:
GHA is correctly running for all the versions of python, but tox didn't
actually run the tests. Fix the config so the tests run and fix the tests.


snuba_sdk/query.py
```diff
@@ -69,63 +67,63 @@ def __post_init__(self) -> None:
         if not isinstance(self.match, Entity):
             raise InvalidQuery("queries must have a valid Entity")
 
-    def _replace(self, field: str, value: Any) -> Query:
+    def _replace(self, field: str, value: Any) -> "Query":
```
lynnagara, line 70: How does this work? What does `"Query"` (as a string) mean?

evanh: I'm still figuring this all out. I'll have to rework large parts of this since dataclasses are not supported in 3.6. This was a stack overflow solution that I was hoping would be a quick fix.

evanh: OK so in 3.6 you can use this string annotation to make a forward reference for mypy. This is replaced in 3.7+ with ` from __future__ import annotations`.

fpacifici (approved):

Makefile
```diff
@@ -13,8 +13,10 @@ help:
 	@false
 
 .venv:
-	virtualenv -ppython3.9 $(VENV_PATH)
-	$(VENV_PATH)/bin/pip install tox
+	virtualenv -ppython3.6 $(VENV_PATH)
+	. $(VENV_PATH)/bin/activate
+	$(VENV_PATH)/bin/pip install -r test-requirements.txt
+	$(VENV_PATH)/bin/pip install -r linter-requirements.txt
```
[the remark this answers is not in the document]
evanh: tox was doing it before. It still is if you run `make test-all` but I pulled these out for easier local testing.

PR #18, opened 2021-02-02
feat(flags) Add missing legacy flags

evanh:
Add sample, consistent, turbo and debug flag support to the SDK.

### DIFF
--- a/snuba_sdk/legacy.py
+++ b/snuba_sdk/legacy.py
@@ -48,7 +48,8 @@ def json_to_snql(body: Mapping[str, Any], entity: str) -> Query:
         return value
 
     dataset = body.get("dataset", "")
-    query = Query(dataset, Entity(entity))
+    sample = body.get("sample")
+    query = Query(dataset, Entity(entity, sample))
 
     selected_columns = list(map(to_exp, body.get("selected_columns", [])))
     for a in body.get("aggregations", []):
@@ -134,10 +135,17 @@ def json_to_snql(body: Mapping[str, Any], entity: str) -> Query:
         limit, name = limitby
         query = query.set_limitby(LimitBy(Column(name), int(limit)))
 
-    extras = ("limit", "offset", "granularity", "totals")
+    extras = (
+        "limit",
+        "offset",
+        "granularity",
+        "totals",
+        "consistent",
+        "turbo",
+        "debug",
+    )
     for extra in extras:
         if body.get(extra) is not None:
             query = getattr(query, f"set_{extra}")(body.get(extra))
 
-    # TODO: Sample clause
     return query
--- a/snuba_sdk/query.py
+++ b/snuba_sdk/query.py
@@ -53,10 +53,10 @@ class Query:
     limit: Optional[Limit] = None
     offset: Optional[Offset] = None
     granularity: Optional[Granularity] = None
-    totals: Optional[Totals] = None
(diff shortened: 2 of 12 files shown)


snuba_sdk/legacy.py
```diff
@@ -134,10 +135,17 @@ def to_scalar(value: Any) -> Any:
         limit, name = limitby
         query = query.set_limitby(LimitBy(Column(name), int(limit)))
 
-    extras = ("limit", "offset", "granularity", "totals")
+    extras = (
+        "limit",
+        "offset",
+        "granularity",
+        "totals",
```
fpacifici, line 142: How does TOTALS work in a joined query ?

snuba_sdk/legacy.py
```diff
@@ -134,10 +135,17 @@ def to_scalar(value: Any) -> Any:
         limit, name = limitby
         query = query.set_limitby(LimitBy(Column(name), int(limit)))
 
-    extras = ("limit", "offset", "granularity", "totals")
+    extras = (
+        "limit",
+        "offset",
+        "granularity",
```
fpacifici, line 141: In the legacy system, the granularity is set by the timeseries extension.
https://github.com/getsentry/snuba/blob/master/snuba/query/timeseries_extension.py#L87

Could you confirm this is performed by the snql parser now that the extension is not executed anymore ?

snuba_sdk/query.py
```diff
@@ -51,6 +54,9 @@ class Query:
     offset: Optional[Offset] = None
     granularity: Optional[Granularity] = None
     totals: Optional[Totals] = None
+    consistent: Optional[Consistent] = None
+    turbo: Optional[Turbo] = None
+    debug: Optional[Debug] = None
```
fpacifici, line 59: Since these are boolean flags they are either True or False. Do we need to make them optional ? That adds a third state (None) that we have to deal with every time we read the flag. IT seems they could have a default set to False. 
The flag object is immutable so you have no problems with mutable default (if they were mutable https://docs.python.org/3/library/dataclasses.html#default-factory-functions)

snuba_sdk/query_visitors.py
```diff
@@ -171,14 +199,32 @@ def _visit_totals(self, totals: Optional[Totals]) -> str:
             return f"TOTALS {self.translator.visit(totals)}"
         return ""
 
+    def _visit_consistent(self, consistent: Optional[Consistent]) -> str:
+        return str(consistent.value) if consistent is not None else ""
```
fpacifici, line 203: This is what I meant before. We should either provide a consistent flag if it is true or no consistent flag at all. Adding a third state adds complexity that does not seems needed.

snuba_sdk/query.py
evanh: I see. I should have looked at the JSON schema definition, I didn't realize they were being defaulted there. I'll default them to False in the SDK as well.

snuba_sdk/legacy.py
evanh: Yes the SnQL parser sets it on the query. https://github.com/getsentry/snuba/blob/master/snuba/query/snql/parser.py#L206

snuba_sdk/legacy.py
evanh: From reading it seems like it work the same as a normal query.

fpacifici (approved):

PR #19, opened 2021-02-04
feat(curried) Add curried functions

evanh:
Add CurriedFunctions, which initializers as well as parameters, and are
formatted as `function(initializers)(parameters)`. These are accepted all the
same places a Function is allowed. A Function is now a CurriedFunction with no
initializers. This also makes the parameters optional, which is reflected by not
encoding the parameters if they are None.

There is a chunk of tech debt added here: the legacy API doesn't support curried
functions, they are sent as raw strings e.g. ["quantile(0.5)", "duration", None]
On top of this some users of the legacy API are sending functions in this form
as well e.g. ["apdex(300, duration)", None, None]. Snuba uses a regex grammar to
parse these strings, and it seemed overkill to port that to this SDK just for
this legacy use case. So instead the raw strings are passed through as is. Once
the legacy code is deprecated we can restrict this again.

### DIFF
--- a/snuba_sdk/expressions.py
+++ b/snuba_sdk/expressions.py
@@ -1,6 +1,6 @@
 import re
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import date, datetime
 from enum import Enum
 from typing import Any, List, Optional, Sequence, Set, Union
@@ -50,9 +50,11 @@ class InvalidArray(Exception):
         )
 
 
-def is_scalar(value: Any) -> bool:
+def is_scalar(value: Any, literal: bool = False) -> bool:
     if isinstance(value, tuple(Scalar)):
         return True
+    elif literal:  # Bail if we don't include arrays/tuples
+        return False
     elif isinstance(value, tuple):
         if not all(is_scalar(v) for v in value):
             raise InvalidExpression("tuple must contain only scalar values")
@@ -66,7 +68,13 @@ def is_scalar(value: Any) -> bool:
     return False
 
 
-column_name_re = re.compile(r"^[a-zA-Z][a-zA-Z_.]+$")
+column_name_re = re.compile(r"^[a-zA-Z](\w|\.)+$")
+# In theory the function matcher should be the same as the column one.
+# However legacy API sends curried functions as raw strings, and it
+# wasn't worth it to import an entire parsing grammar into the SDK
+# just to accomodate that one case. Instead, allow it for now and
+# once that use case is eliminated we can remove this.
+function_name_re = re.compile(r"^[a-zA-Z](\w|[().,]| |\[|\])+$")
 
 
 def _validate_int_literal(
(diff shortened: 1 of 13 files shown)


snuba_sdk/expressions.py
```diff
@@ -50,9 +50,11 @@ def __init__(self, value: List[Any]) -> None:
         )
 
 
-def is_scalar(value: Any) -> bool:
+def is_scalar(value: Any, literal: bool = False) -> bool:
```
fpacifici, line 53: It seems there should be just a separate is_literal method.

snuba_sdk/expressions.py
```diff
@@ -66,7 +68,13 @@ def is_scalar(value: Any) -> bool:
...
 
 
-column_name_re = re.compile(r"^[a-zA-Z][a-zA-Z_.]+$")
+column_name_re = re.compile(r"^[a-zA-Z](\w|\.)+$")
+# In theory the function matcher should be the same as the column one.
+# However legacy API sends curried functions as raw strings, and it
+# wasn't worth it to import an entire parsing grammar into the SDK
+# just to accomodate that one case. Instead, allow it for now and
+# once that use case is eliminated we can remove this.
+function_name_re = re.compile(r"^[a-zA-Z](\w|[().,]| |\[|\])+$")
```
fpacifici, line 77: What would the incentive to remove this use case be ? If the clients will find it convenient they will not stop using it.

snuba_sdk/expressions.py
```diff
@@ -155,18 +163,25 @@ def validate(self) -> None:
 
 
 @dataclass(frozen=True)
-class Function(Expression):
+class CurriedFunction(Expression):
     function: str
-    parameters: Sequence[Union[ScalarType, Column, "Function"]]
+    initializers: Optional[Sequence[Union[ScalarLiteralType, Column]]] = None
```
fpacifici, line 168: Why cannot the initializer contain Function calls itself? Is it something clickhouse does not support or a constraint we are imposing. If it is the second, what's the reason?

snuba_sdk/legacy.py
```diff
@@ -18,48 +18,50 @@ def parse_datetime(date_str: str) -> datetime:
...
+def parse_exp(value: Any) -> Any:
+    if isinstance(value, str):
+        return Column(value)
+    if not isinstance(value, list):
+        return parse_scalar(value)
+
+    alias = value[2] if len(value) > 2 else None
 
+    if value[0].endswith("()") and not value[1]:
+        return Function(value[0].strip("()"), [], alias)
```
fpacifici, line 46: What is this case for ?

fpacifici (commented):
A few clarifications otherwise it seems ok

snuba_sdk/expressions.py
evanh: Well right now the clients have to use it in order to do things like parametric functions. Once we have SnQL everywhere, they no longer have to use it and we can deprecate it completely.

snuba_sdk/expressions.py
evanh: It is self imposed, and the reason is two fold: firstly, there is no use case for it, secondly, we don't support it in Snuba.

snuba_sdk/legacy.py
evanh: Sometimes clients are sending `["count()", None, ...]`.

fpacifici (approved):

PR #21, opened 2021-02-22
feat(docs) Add sphinx to SDK to generate docs

evanh:
Right now these are autogenerated stubs. There will be a follow up PR to host
these documents somewhere.

Screen Shot 2021-02-22 at 1 52 32 PM


docs/conf.py
```diff
@@ -0,0 +1,85 @@
...
+sys.path.insert(0, os.path.abspath(".."))
+
+# -- Project information -----------------------------------------------------
+
+project = "snuba-sdk"
+copyright = "2021, Sentry Team and Contributors"
+author = "Sentry Team and Contributors"
+
+# The full version, including alpha/beta/rc tags
+release = "0.0.5"
```
fpacifici, line 30: Is this supposed to be the version of the sdk? Can we have it centralized in a file and imported everywhere it is used?

fpacifici (approved):
Could you please see my comment on the version before merging.

docs/conf.py
evanh: I think ultimately this is managed automatically. The code that generates the release also automatically updates the release number in the code base.

PR #22, opened 2021-02-24
fix(subscriptables) Allow subscriptables in the columns

evanh:
Allow subscriptables (column[key]) in the Column class. If a Column is a
subscriptable, also populate two helper fields column and key.

### DIFF
--- a/snuba_sdk/expressions.py
+++ b/snuba_sdk/expressions.py
@@ -173,7 +173,7 @@ class Column(Expression):
     """
 
     name: str
-    column: Optional[str] = field(init=False, default=None)
+    subscriptable: Optional[str] = field(init=False, default=None)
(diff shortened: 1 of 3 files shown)


snuba_sdk/expressions.py
```diff
@@ -155,7 +157,24 @@ class Debug(BooleanFlag):
...
+    :param name: The column name.
+    :type name: str
+
+    :raises InvalidExpression: If the column name is not a string or has an
+        invalid format.
+
+    """
+
     name: str
+    column: Optional[str] = field(init=False, default=None)
```
fpacifici, line 176: The difference between column and name is kind of confusing. I would call column and keys something like `subscriptable` and `key`

fpacifici (approved):

PR #23, opened 2021-02-24
feat(docs) Automate doc generation

evanh:
Add a GHA to auto generate the sphinx docs. Right now this is on every PR but
I'll change it once I'm confident this actually works.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,36 @@
+name: API Docs
+
+on:
+  push:
+    branches:
+      - main
+    pull_request:
+
+jobs:
+  docs:
+    name: Sphinx
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+
+      - name: Build docs
+        run: |
+          pip install virtualenv
+          make apidocs
+
+      - uses: peaceiris/actions-gh-pages@v3.7.3
```
fpacifici, line 25: Do we have a list of vetted actions we are allowed to use in our repos?

evanh: No but I got this directly from the discuss-docs channel. It is being used in a bunch of our current repos.

fpacifici (approved):

PR #26, opened 2021-02-26
feat(boolean) Add BooleanConditions to the SDK

evanh:
Add a BooleanCondition expression that takes a boolean operator (AND/OR) and a
sequence of other conditions and performs that boolean operator.

They are translated as (cond OP cond ...) for each cond in the conditions.


lynnagara:
I was kind of surprised to see booleans treated more like conditions rather than functions. Perhaps the distinction is arbitrary but in my mind they are dissimilar to conditions since they do not do any filtering of results per se. After all isn't `a AND b` just an alternative syntax for the function expression `and(a,b)`? Also does this then imply that a boolean is only valid in the WHERE/HAVING part of the query, and not, say, in SELECT?

evanh:
There is nothing preventing a user from adding `FunctionCall("and", [...])` to any other clause that supports `FunctionCall`. You're right that this is basically syntactic sugar over a function call. It makes the code a little easier to read and a little easier for us to validate.

```
FunctionCall("equals", [FunctionCall("or", [...])], 1)
```
vs.
```
BooleanCondition(BooleanOp.OR, [...])
```

snuba_sdk/conditions.py
```diff
@@ -59,3 +59,35 @@ def validate(self) -> None:
...
+
+class BooleanOp(Enum):
+    AND = "AND"
+    OR = "OR"
+
+
+@dataclass(frozen=True)
+class BooleanCondition(Expression):
+    op: BooleanOp
+    conditions: Sequence[Union["BooleanCondition", Condition]]
```
fpacifici, line 72: I like the idea of simplifying the syntax for conditions. I wonder if this needs to be a first class expression type (which has its own class and thus it is visible to the visitor).
Have you consider the alternative of building the DSL on top of the existing expressions like `and(condition, condition, etc.)` which would be a function that produces a `FunctionCall("AND", etc.)`. 
Which keeps the first class expressions to deal with simpler (simpler visitors).
If yes, what was the trade off? And what was the reasoning to make BooleanCondition a first class expression?

To be clear, the alternative I am bringing up would still be part of the sdk repo, just not a dedicated expression type.

tests/test_condition.py
```diff
@@ -103,7 +104,144 @@ def verify() -> None:
...
+            ],
+        ),
+        BooleanCondition(
+            BooleanOp.AND,
+            [
+                Condition(Column("a"), Op.EQ, 1),
+                Condition(Column("b"), Op.EQ, 2),
+                Condition(Column("c"), Op.EQ, 3),
+            ],
+        ),
```
fpacifici, line 143: If we go this way I would make BooleanCondition abstract and provide two subclasses or helper functions:
`And` and `Or`. Imagine something like: `and(condition, condition, condition ....)`

lynnagara:
If we consider BooleanCondition to be just syntactic sugar for the function expressions `and()` and `or()`, should we expect them to be valid everywhere that the other functions are valid? From what I can tell, BooleanConditions are only allowed in the where and having clauses. This means a user might have to combine the two syntaxes in different parts of the same query because they can't reuse their BooleanCondition in say a SELECT statement, which may not be very user friendly. (Same applies for any other conditions, not just these boolean ones I suppose.)

tests/test_condition.py
evanh: I like this idea. I'll add it.

evanh:
> And what was the reasoning to make BooleanCondition a first class expression?

@fpacifici The reason I made [Boolean]Condition its own class trickles back to how SnQL conditions are formatted. While Clickhouse allows prefix and infix conditions, the SnQL spec expects only infix conditions. That means any condition function would need to wrapped with "= 1" when it's translated to a SnQL query. While that is certainly valid, it's not very readable and makes it that much more difficult to determine what a "top level" condition is. So I added a separate expression that has its own translation logic to create the infix string that SnQL expects. I could certainly add the ability to be able to freely switch between infix and prefix in the same way Clickhouse does, but that seems like a lot of work for a product no one is actually using.

> should we expect them to be valid everywhere that the other functions are valid?
@lynnagara This is related to the answer above. While the answer is theoretically yes (again, Clickhouse allows it) the current SnQL parser would not correctly parse AND/OR expressions anywhere other than WHERE/HAVING. Supporting this is certainly possible but I'd like to get some user feedback on it before spending too much time on it.

fpacifici:
> That means any condition function would need to wrapped with "= 1" when it's translated to a SnQL query. 

Makes sense to me

snuba_sdk/query.py
```diff
@@ -47,8 +47,8 @@ class Query:
     match: Entity
     select: Optional[List[Union[Column, CurriedFunction, Function]]] = None
     groupby: Optional[List[Union[Column, CurriedFunction, Function]]] = None
-    where: Optional[List[Condition]] = None
-    having: Optional[List[Condition]] = None
+    where: Optional[List[Union[BooleanCondition, Condition]]] = None
+    having: Optional[List[Union[BooleanCondition, Condition]]] = None
```
fpacifici, line 51: I think you may want to make this a `Condition`. If there is an `And` condition I don't see a strong need to also allow a List of `Condition` which mean the same thing by introducing some minimal overhead.
IT would make processing and readability easier as people would not have to remember the meaning of a sequence of conditions.

fpacifici (commented):
I am ok with this. Will let @lyn address your answer to her comment.

snuba_sdk/query.py
evanh: You mean to make this the same as the Snuba AST? Where there is one base condition that holds all the others?

fpacifici: yes

lynnagara (commented):
So to recap, this PR has the additional goal of simplifying the clickhouse query by avoiding the =1 on conditions, and we solve that though using switching to infix notation in WHERE/HAVING only. This covers most of our use cases, but infix is not generally supported, and =1 still being applied everywhere else. Is that accurate?

> > That means any condition function would need to wrapped with "= 1" when it's translated to a SnQL query.
> 
> Makes sense to me

I'm not sure I get why we need to wrap expressions in "=1". This seems like an attempt to make invalid queries valid. If that's true, why wouldn't we just error when we get an invalid query rather than always attempt to coerce it to a valid but different thing than may be intended.

snuba_sdk/query.py
evanh: That will unnecessarily clutter up the interface I think, since you'll have to use `And`/`Or` even if you only have one condition. I'm not sure why it would be more readable.

lynnagara (approved):

PR #28, opened 2021-03-02
feat(subquery) Add subquery support to SDK

evanh:
Add the ability to pass another Query in the match clause of a query. Also add
validation for the inner query and to ensure the outer query is only using
fields returned by the inner query. Added a new visitor that can be used to find
all the expressions of a certain type in a query.

### DIFF
--- a/snuba_sdk/query_visitors.py
+++ b/snuba_sdk/query_visitors.py
@@ -1,12 +1,13 @@
 import json
 from abc import ABC, abstractmethod
 from typing import (
+    Any,
     Generic,
     Mapping,
     MutableMapping,
     Optional,
     Sequence,
-    TYPE_CHECKING,
+    Set,
     TypeVar,
     Union,
 )
@@ -28,12 +29,11 @@ from snuba_sdk.expressions import (
     Totals,
     Turbo,
 )
-from snuba_sdk.visitors import Translation
+from snuba_sdk.visitors import ExpressionFinder, Translation
 
-if TYPE_CHECKING:
-    # Import the module due to sphinx autodoc problems
-    # https://github.com/agronholm/sphinx-autodoc-typehints#dealing-with-circular-imports
-    from snuba_sdk import query
+# Import the module due to sphinx autodoc problems
+# https://github.com/agronholm/sphinx-autodoc-typehints#dealing-with-circular-imports
+from snuba_sdk import query as main
 
 
 class InvalidQuery(Exception):
@@ -44,7 +44,7 @@ QVisited = TypeVar("QVisited")
 
 
 class QueryVisitor(ABC, Generic[QVisited]):
-    def visit(self, query: "query.Query") -> QVisited:
+    def visit(self, query: "main.Query") -> QVisited:
         fields = query.get_fields()
         returns = {}
         for field in fields:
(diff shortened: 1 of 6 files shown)


snuba_sdk/visitors.py
```diff
@@ -228,3 +228,82 @@ def _visit_turbo(self, turbo: Turbo) -> str:
...
+
+    def _visit_turbo(self, turbo: Turbo) -> Set[Expression]:
+        if isinstance(turbo, self.exp_type):
+            return set([turbo])
+        return set()
+
+    def _visit_debug(self, debug: Debug) -> Set[Expression]:
+        if isinstance(debug, self.exp_type):
+            return set([debug])
+        return set()
```
fpacifici, line 309: Aren't these boolean flags ? Do we need to find them with an expression finder ?

fpacifici (commented):
A couple of clarifications.

snuba_sdk/query_visitors.py
```diff
@@ -128,20 +128,22 @@ def _visit_debug(self, debug: Debug) -> QVisited:
...
 
 class Printer(QueryVisitor[str]):
-    def __init__(self, pretty: bool = False) -> None:
+    def __init__(self, pretty: bool = False, is_inner: bool = False) -> None:
         self.translator = Translation()
         self.pretty = pretty
+        self.is_inner = is_inner
 
-    def _combine(self, query: "query.Query", returns: Mapping[str, str]) -> str:
+    def _combine(self, query: "main.Query", returns: Mapping[str, str]) -> str:
```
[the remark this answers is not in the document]
evanh: The arguments to the function is also called `query` so it was clashing with the package name.

snuba_sdk/visitors.py
evanh: Probably not but why not add the functionality.

snuba_sdk/query_visitors.py
```diff
@@ -244,10 +256,109 @@ def _combine(self, query: "query.Query", returns: Mapping[str, str]) -> str:
...
+            self.visit(query.match)
+
+            outer_exps = self.column_finder.visit(query)
+            inner_match = set()
+            assert query.match.select is not None
+            for s in query.match.select:
+                if isinstance(s, CurriedFunction):
+                    inner_match.add(s.alias)
+                elif isinstance(s, Column):
+                    inner_match.add(s.name)
```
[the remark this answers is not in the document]
evanh: That's an interesting idea. If this pattern becomes more common (which it might with Join validation) that would simplify things.

snuba_sdk/visitors.py
```diff
@@ -228,3 +228,82 @@ def _visit_turbo(self, turbo: Turbo) -> str:
 
     def _visit_debug(self, debug: Debug) -> str:
         return str(debug)
+
+
+class ExpressionFinder(ExpressionVisitor[Set[Expression]]):
+    def __init__(self, exp_type: Any) -> None:
```
fpacifici, line 234: Can this type be restricted or would this be a massive Union ?

fpacifici (approved):

snuba_sdk/visitors.py
evanh: It would be massive because it's the type of any Expression.

PR #30, opened 2021-03-03
fix(legacy) Handle arrayjoin validation

evanh:
Arrayjoin queries are handled differently in the groupby clause. Handle the
difference in the validation.


snuba_sdk/query_visitors.py
```diff
@@ -288,8 +288,23 @@ def _combine(self, query: "query.Query", returns: Mapping[str, None]) -> None:
...
-                    raise InvalidQuery(f"{group_exp} missing from the groupby")
+                # Array joins work a little differently. In this case the entire arrayjoin
+                # expression shouldn't be in the groupby, just the column that is being
+                # array joined
+                if (
+                    isinstance(group_exp, Function)
+                    and group_exp.function == "arrayJoin"
+                ):
+                    assert group_exp.parameters and isinstance(
+                        group_exp.parameters[0], Column
```
fpacifici, line 299: Are you sure we cannot arrayJoin over a complex expression? `arrayJoin(f(x))`? It should still work.

evanh: This is old. I'll update.

snuba_sdk/legacy.py
```diff
@@ -15,27 +15,43 @@
...
+            return escaped
 
     return value
 
 
 def parse_exp(value: Any) -> Any:
     if isinstance(value, str):
+        if not value or value.startswith("'"):
+            return value.strip("'")
+
```
fpacifici, line 54: I am not sure I get what this impact arrayJoin validations.

evanh: These changes are to properly handle the test case I added, which is from the snuba discover tests.

fpacifici (approved):

PR #33, opened 2021-03-08
feat(column-entity) Add an alias to entity and column 

evanh:
Add an alias field to an entity, and add an entity field to a column. This is in
preparation for adding JOIN support. Having Column dependent on Entity forced a
code restructure which was huge as everything uses Column. Move all complex
expressions into their own modules.

### DIFF
--- a/snuba_sdk/entity.py
+++ b/snuba_sdk/entity.py
@@ -15,12 +15,13 @@ class InvalidEntity(Exception):
 @dataclass(frozen=True)
 class Entity(Expression):
     name: str
+    alias: Optional[str] = None
     sample: Optional[Union[int, float]] = None
 
     def validate(self) -> None:
         # TODO: There should be a whitelist of entity names at some point
         if not isinstance(self.name, str) or not entity_name_re.match(self.name):
-            raise InvalidEntity(f"{self.name} is not a valid entity name")
+            raise InvalidEntity(f"'{self.name}' is not a valid entity name")
 
         if self.sample is not None:
             if not isinstance(self.sample, (int, float)):
@@ -35,3 +36,7 @@ class Entity(Expression):
             elif isinstance(self.sample, int):
                 if self.sample < 1:
                     raise InvalidEntity("int samples must be at least 1 (# of rows)")
+
+        if self.alias is not None:
+            if not isinstance(self.alias, str) or not self.alias:
+                raise InvalidEntity(f"'{self.alias}' is not a valid alias")
--- a/snuba_sdk/query_visitors.py
+++ b/snuba_sdk/query_visitors.py
@@ -339,8 +339,8 @@ class Validator(QueryVisitor[None]):
         # - Must have certain conditions (project, timestamp, organization etc.)
         ## SUBQUERIES
         # - outer query must only reference columns from inner query, and reference by alias
+        all_columns = self.column_finder.visit(query)
         if isinstance(query.match, main.Query):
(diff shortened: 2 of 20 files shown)


snuba_sdk/entity.py
```diff
@@ -15,12 +15,13 @@ class InvalidEntity(Exception):
 @dataclass(frozen=True)
 class Entity(Expression):
     name: str
+    alias: Optional[str] = None
```
evanh, line 18: This is another functional change

snuba_sdk/query_visitors.py
```diff
@@ -341,10 +339,7 @@ def _combine(self, query: "main.Query", returns: Mapping[str, None]) -> None:
         # - Must have certain conditions (project, timestamp, organization etc.)
         ## SUBQUERIES
         # - outer query must only reference columns from inner query, and reference by alias
-        # - inner query must be valid
         if isinstance(query.match, main.Query):
-            self.visit(query.match)
```
fpacifici, line 346: Why this change ?
Moreover, shouldn't we validate that, if the query is not a join entity alias must not be used?

evanh: This is doing duplicate work. The combine function is called after the `_visit_match` function so the query will already be validated.

evanh: The validation comment is valid, I'll check that.

evanh: So instead of adding validation I decided to add a flag to the translator that will omit the entity aliases for non-join queries. The reason why I did this was to avoid the end user having to have different entity/column values (mostly entity) when they are building join vs. non-join queries. This way if the end user wants to have a static set of Entities that get reused in different queries, they only need to have one set no matter which query they are building.

snuba_sdk/visitors.py
```diff
@@ -153,8 +151,16 @@ def _visit_debug(self, debug: Debug) -> TVisited:
 
 
 class Translation(ExpressionVisitor[str]):
+    def __init__(self, use_entity_aliases: bool = False):
+        # Eventually JOINs will set this to True, but single entity/sub queries
+        # don't support entity aliases.
+        self.use_entity_aliases = use_entity_aliases
```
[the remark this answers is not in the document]
evanh: This is probably better addressed in the Join PR (https://github.com/getsentry/snuba-sdk/pull/35).

> Isn't this going to miss the validation rule that all columns in the joins need to reference a table alias?

No in the Join PR I add explicit validation for that.

> This way we should be able to validate different types of queries cleanly and without letting a simple query get into a state that is undefined for it or the join query to get into an invalid state.

I'm not sure how this would prevent that now. Queries in general (not just Join queries) can already exist in an invalid state, since they can be built up over time. They are not expected to be valid until `query.validate()` is called. I did investigate doing the split you are talking about, but ultimately the difference in validation ended up being just a few extra lines of code.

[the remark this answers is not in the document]
fpacifici: Sorry I missed the gap between this file and the previous one. I thought this was the query translator while it was the expression one. Yeah my suggestion would not help with the validation rule for joins.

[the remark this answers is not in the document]
fpacifici: > I did investigate doing the split you are talking about, but ultimately the difference in validation ended up being just a few extra lines of code.

Ok.

fpacifici (approved):

PR #35, opened 2021-03-09
feat(join) Add Join/Relationship classes and query support

evanh:
Depends on #33 

Add Join class that contains an array of Relationships, each of which describes
two entities and the join relationship between them. A query can now have a Join
as  the match value. This also adds a restriction that all columns in a Join
query must be qualified with an entity that also exists in the Join.

### DIFF
--- a/snuba_sdk/join.py
+++ b/snuba_sdk/join.py
@@ -1,0 +1,54 @@
+from dataclasses import dataclass
+from typing import Any, MutableMapping, Sequence, Tuple
+
+from snuba_sdk.entity import Entity
+from snuba_sdk.expressions import Expression, InvalidExpression
+
+
+@dataclass(frozen=True)
+class Relationship(Expression):
+    lhs: Entity
+    name: str
+    rhs: Entity
+
+    def validate(self) -> None:
+        def valid_entity(e: Any) -> None:
+            if not isinstance(e, Entity):
+                raise InvalidExpression(f"'{e}' must be an Entity")
+            elif e.alias is None:
+                raise InvalidExpression(f"{e} must have a valid alias")
+
+        valid_entity(self.lhs)
+        valid_entity(self.rhs)
+
+        if not isinstance(self.name, str) or not self.name:
+            raise InvalidExpression(f"'{self.name}' is not a valid relationship name")
+
+
+@dataclass(frozen=True)
+class Join(Expression):
+    relationships: Sequence[Relationship]
+
+    def get_alias_mappings(self) -> Sequence[Tuple[str, str]]:
+        aliases = []
+        for rel in self.relationships:
+            if rel.lhs.alias is not None:
+                aliases.append((rel.lhs.alias, rel.lhs.name))
+            if rel.rhs.alias is not None:
+                aliases.append((rel.rhs.alias, rel.rhs.name))
+
+        return aliases
+
+    def validate(self) -> None:
+        if not isinstance(self.relationships, (list, tuple)) or not self.relationships:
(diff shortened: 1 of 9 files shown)


snuba_sdk/query_visitors.py
```diff
@@ -354,6 +359,23 @@ def _combine(self, query: "main.Query", returns: Mapping[str, None]) -> None:
                     raise InvalidQuery(
                         f"outer query is referencing column {c.name} that does not exist in subquery"
                     )
+        elif isinstance(query.match, Join):
```
fpacifici, line 362: Her I would reiterate my comment I provided in the PR below this one. 
While the difference between simple query, composite and join is just 20 lines each, the issue is the cognitive overload of having three responsibilities in this same class, which in this case are three different modes of functioning. 
Being in the same class the separation between these functions is less clear than it could be.
By splitting them instead, the structure of the code itself immediately tells who takes care of what and that there are three different ways to validate. This is useful for people who read this code for the first time and are less familiar with the code base.

tests/test_join.py
```diff
@@ -0,0 +1,249 @@
...
+from snuba_sdk.orderby import Direction, OrderBy
+from snuba_sdk.query import Query
+from snuba_sdk.query_visitors import InvalidQuery
+
+tests = [
+    pytest.param(
+        Query(
+            "discover",
+            Join([Relationship(Entity("events", "e"), "has", Entity("spans", "s"))]),
+        )
```
fpacifici, line 20: sadly spans are not a thing anymore. Up to you, but you may want to reference groupedmessage instead so people do not get confused.

fpacifici (commented):
A few comments but no conceptual issue

snuba_sdk/join.py
```diff
@@ -0,0 +1,54 @@
...
+
+        if not isinstance(self.name, str) or not self.name:
+            raise InvalidExpression(f"'{self.name}' is not a valid relationship name")
+
+
+@dataclass(frozen=True)
+class Join(Expression):
+    relationships: Sequence[Relationship]
+
+    def get_alias_mappings(self) -> Sequence[Tuple[str, str]]:
```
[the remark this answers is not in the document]
evanh: Hmm this might have been a holdover. This was originally a Sequence[str] so that I could more easily track duplicates but now I think I don't need that.

tests/test_join.py
evanh: Yeah that's a good point.

snuba_sdk/query_visitors.py
evanh: I disagree. I don't think a single if statement is harder to read than multiple separate classes. Particularly because there is only this one small difference between the classes. I think you could make a case that it's easier to read, since all the logic is in one place instead of being spread around.

fpacifici: Mh, let's find a compromise.

I believe you are right tin saying that one single branch in the behavior of a class is not really harmful. It becomes more concerning when we add more. Then the responsibilities of the class increase exponentially. At that point, usually, if we are not super disciplined, we do not refactor the class because it is too complex.

Would you agree with that point? If yes and we commit in doing such refactoring when needed, then there would not be any issue with this PR, and I will accept.

evanh: I definitely agree. And I am definitely open to doing the refactoring. As I mentioned before, I was already considering doing it for this change. Any more divergence would probably warrant a splitting of logic.

fpacifici (approved):

PR #37, opened 2021-03-16
feat(legacy) Have the SDK convert Legacy to maintain SQL equivalence

evanh:
Make a number of small changes/fixes to ensure that when a Legacy API call is
translated to an SDK object, it will ultimately return the same SQL that the
legacy call would have generated.


snuba_sdk/entity.py
```diff
@@ -16,27 +16,38 @@ class InvalidEntity(Exception):
...
+                raise InvalidEntity("sample must be a float")
+            elif self.sample <= 0.0:
+                raise InvalidEntity("samples must be greater than 0.0")
 
         if self.alias is not None:
             if not isinstance(self.alias, str) or not self.alias:
                 raise InvalidEntity(f"'{self.alias}' is not a valid alias")
+
+
+ENTITY_TIME_COLUMNS = {
```
evanh, line 37: Ultimately this should be handled by the users of the SDK, not the SDK itself. For now this is here.

tests/test_legacy.py
```diff
@@ -37,17 +37,17 @@
             "-- CONSISTENT: True",
             "-- DEBUG: True",
             "MATCH (sessions SAMPLE 0.100000)",
-            "SELECT project_id, release, min(max(users) AS max_users) AS min_users",
+            "SELECT min(max(users) AS max_users) AS min_users, project_id, release",
```
evanh, line 40: All of this file is reordering the select columns / conditions to match the order they are processed by the legacy API.

setup.py
```diff
@@ -9,7 +9,7 @@
 """
 
 import os
-from setuptools import setup, find_packages
+from setuptools import setup, find_packages  # type: ignore
```
fpacifici, line 12: what was the issue here ? Mind adding a comment that explains ?

snuba_sdk/entity.py
```diff
@@ -16,27 +16,39 @@ class InvalidEntity(Exception):
...
+            elif self.sample <= 0.0:
+                raise InvalidEntity("samples must be greater than 0.0")
 
         if self.alias is not None:
             if not isinstance(self.alias, str) or not self.alias:
                 raise InvalidEntity(f"'{self.alias}' is not a valid alias")
+
+
+# TODO: This should be handled by the users of the SDK, not the SDK itself.
+ENTITY_TIME_COLUMNS = {
```
fpacifici, line 38: Do you have a plan on how to move these out of the sdk? I think having these hardcoded is a pretty serious issue as it is not going to be required for the migration only. Also we will forget to keep them up to date. 

From the old query language perspective these columns are dataset specific, not entity specific (they do not know entities). What about providing a mapping dataset to timestamp column to the sdk when querying ? We could make it transparent into utils/snuba.py

snuba_sdk/legacy.py
```diff
@@ -66,14 +117,81 @@ def parse_exp(value: Any) -> Any:
     return Function(value[0], children, alias)
 
 
+def parse_automatic_condition(
```
fpacifici, line 120: What about calling this `parse_extension_condition` ? That would make it pretty clear what this is for.

snuba_sdk/legacy.py
```diff
@@ -66,14 +117,81 @@ def parse_exp(value: Any) -> Any:
...
+            values = (values,)
+        else:
+            return Condition(column, Op.EQ, values)
+
+    if isinstance(values, list):
+        rhs: Sequence[Any] = tuple(map(parse_scalar, values))
+        return Condition(column, Op.IN, rhs)
+    elif isinstance(values, tuple):
+        rhs = tuple(map(parse_scalar, values))
+        return Condition(column, Op.IN, rhs)
```
fpacifici, line 149: Am I missing something or both branches do the same thing ?

snuba_sdk/entity.py
evanh: So there are two reasons why this won't get stagnant: firstly, this is only necessary for legacy -> snql conversions. Once that need goes away, so does this. Secondly, I plan to rewrite all of this when I add the data model component to the SDK, so this will disappear at that point as well.

Having Sentry provide the mapping is fine, but that data will have to live in 3 places then: the SDK and Snuba (for tests) as well as Sentry. Also is it worth the added complexity for logic that is slated to be written before the end of the quarter?

setup.py
evanh: Mypy claims setuptools is missing type annotations. Maybe a 3.6 error?

snuba_sdk/legacy.py
evanh: No you're right. These branches originally had different logic and I must have forgotten to collapse them.

snuba_sdk/entity.py
fpacifici: > firstly, this is only necessary for legacy -> snql conversions.

To confirm. We do not plan to add mandatory conditions validation in the sdk, right? 

> Secondly, I plan to rewrite all of this when I add the data model component to the SDK, so this will disappear at that point as well.

This sounds a good plan to me.

fpacifici (approved):

snuba_sdk/entity.py
evanh: > To confirm. We do not plan to add mandatory conditions validation in the sdk, right? 

I do plan to add the concept of mandatory condition validation to the SDK, but it would be up to user to define what the actual conditions should be. Similar to how the Snuba Entity has a required_columns field and knows how to validate that field, but the specifics of which field should be validated is left to the user of that class.

PR #41, opened 2021-04-12
fix(legacy) Handle alias columns in the select

evanh:
The legacy API uses aliases in the select clause for aggregations. Correctly
validate this case.

### DIFF
--- a/snuba_sdk/query_visitors.py
+++ b/snuba_sdk/query_visitors.py
@@ -405,8 +405,10 @@ class Validator(QueryVisitor[None]):
         alias_columns = set()
         for exp in query.select:
             if (
-                isinstance(exp, (CurriedFunction, Function)) and exp.is_aggregate()
-            ) and exp.alias:
+                isinstance(exp, (CurriedFunction, Function))
+                and exp.is_aggregate()
+                and exp.alias
+            ):
                 alias_columns.add(Column(exp.alias))
 
         for exp in query.select:
--- a/snuba_sdk/query_visitors.py
+++ b/snuba_sdk/query_visitors.py
@@ -5,6 +5,7 @@ from typing import (
     Generic,
     Mapping,
     MutableMapping,
+    List,
     Optional,
     Sequence,
     Set,
@@ -396,12 +397,23 @@ class Validator(QueryVisitor[None]):
             raise InvalidQuery("query must have at least one expression in select")
 
         # Non-aggregate expressions must be in the groupby if there is an aggregate
-        non_aggregates = []
+        non_aggregates: List[Union[Column, CurriedFunction, Function]] = []
         has_aggregates = False
+
+        # Legacy queries use aliases as Columns, so we need to compile all the possible
+        # aliases to use in the checks.
+        alias_columns = set()
         for exp in query.select:
             if (
-                not isinstance(exp, (CurriedFunction, Function))
-                or not exp.is_aggregate()
+                isinstance(exp, (CurriedFunction, Function)) and exp.is_aggregate()
+            ) and exp.alias:
+                alias_columns.add(Column(exp.alias))
+
+        for exp in query.select:
+            if isinstance(exp, Column) and exp not in alias_columns:
+                non_aggregates.append(exp)
+            elif (
+                isinstance(exp, (CurriedFunction, Function)) and not exp.is_aggregate()
             ):
                 non_aggregates.append(exp)
             else:
--- a/snuba_sdk/query_visitors.py
+++ b/snuba_sdk/query_visitors.py
@@ -400,8 +400,10 @@ class Validator(QueryVisitor[None]):
         non_aggregates: List[Union[Column, CurriedFunction, Function]] = []
         has_aggregates = False
 
-        # Legacy queries use aliases as Columns, so we need to compile all the possible
-        # aliases to use in the checks.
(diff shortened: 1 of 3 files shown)


snuba_sdk/query_visitors.py
```diff
@@ -396,12 +397,23 @@ def _combine(self, query: "main.Query", returns: Mapping[str, None]) -> None:
             raise InvalidQuery("query must have at least one expression in select")
 
         # Non-aggregate expressions must be in the groupby if there is an aggregate
-        non_aggregates = []
+        non_aggregates: List[Union[Column, CurriedFunction, Function]] = []
```
fpacifici, line 400: Is this change related with this warning `"groupby must be included if there are aggregations in the select" `  from prod ?

snuba_sdk/query_visitors.py
```diff
@@ -396,12 +397,23 @@ def _combine(self, query: "main.Query", returns: Mapping[str, None]) -> None:
...
+
+        # Legacy queries use aliases as Columns, so we need to compile all the possible
+        # aliases to use in the checks.
+        alias_columns = set()
         for exp in query.select:
             if (
-                not isinstance(exp, (CurriedFunction, Function))
-                or not exp.is_aggregate()
+                isinstance(exp, (CurriedFunction, Function)) and exp.is_aggregate()
+            ) and exp.alias:
```
fpacifici, line 409: I think this is not properly formatted.
```
   if isinstance(....bla....) and exp.is_aggregate() and exp.alias
```

snuba_sdk/query_visitors.py
```diff
@@ -396,12 +397,23 @@ def _combine(self, query: "main.Query", returns: Mapping[str, None]) -> None:
             raise InvalidQuery("query must have at least one expression in select")
 
         # Non-aggregate expressions must be in the groupby if there is an aggregate
-        non_aggregates = []
+        non_aggregates: List[Union[Column, CurriedFunction, Function]] = []
         has_aggregates = False
+
+        # Legacy queries use aliases as Columns, so we need to compile all the possible
```
fpacifici, line 403: Not sure I get what use case this refers to. 
For `use aliases as Columns` do you mean that we would find identifier in the select clause of the legacy query that are transformed into `Column` by the legacy to snql translator?

If yes, what do we need the additional condition `exp.alias:` for at line 409 ?

fpacifici (commented):
I think I am missing a bit of context on what this PR would fix

snuba_sdk/query_visitors.py
evanh: The best example is the test case I added. Some legacy queries have expressions in the select clause and the aggregation clause, e.g. `{"selected_columns": ["x"], "aggregations": [["fn", ..., "x"]]...`. There is no groupby necessary in this query since there is only the aggregation selected, but the SDK would incorrectly think that `"x"` was a raw column and complain that it wasn't in the groupby clause.

snuba_sdk/query_visitors.py
evanh: Yes.

snuba_sdk/query_visitors.py
fpacifici: It seems still improperly formatted. Is black doing this ?

snuba_sdk/query_visitors.py
fpacifici: Still I don't get your condition.
Line 410 adds a column to `alias_columns` if the expression is a function, it is aggregate and it has an alias. 
Isn't this something that could be checked in the same loop at line 412?
Or is the select looking like this after the legacy translator resolved the aggregation?
```
["first_session_started", "last_session_started", "min(started)", ""max(started)" ]
```
If this is the case, could you please comment with an explanation of this use case. Reading the test case does not show what the reason for this behavior is.

Also isn't this a general problem we can encounter in validation potentially anywhere by having aliases in the expressions ?

snuba_sdk/query_visitors.py
evanh: Sorry forgot to push the change.

snuba_sdk/query_visitors.py
evanh: > Or is the select looking like this after the legacy translator resolved the aggregation?

Yes that's what the select looks like. Then the validation fails because it thinks `"first_session_started"` and `"last_session_started"` should be in the groupby.

In order to check that, I need to know the aliases of all the aggregates so in the second loop I can see if a given select expression is an alias of an aggregate or a column that should be in the groupby.

As for a general problem, the answer is "yes, but it hackily works right now". It only becomes an issue with these validators because they need to understand the difference between a Column and "something that looks like a Column but is actually an alias". Nowhere else in the code does that distinction matter (right now). I've added a tech debt task to better handle these situations and make this more explicit in the SDK, but that is not something I intend to tackle in this PR.

fpacifici: Ok, could you please add a comment explaining the corner case we are dealing with. 

Also for the general case, can we just get rid of aliases entirely once the legacy translator is gone ?

evanh: That's a discussion point. That was my original intention when I started the SDK but I'm starting to wonder if we should support them for usability reasons.

fpacifici (approved):

PR #48, opened 2021-04-23
fix(groupby) Handle legacy queries with nested aliases of aggregates

evanh:
This fix is to handle legacy queries where some of the expressions are functions
on aliases of other aggregate functions, e.g. `avg(x) AS avg_x, minus(avg_x, 2)`
When checking to see if the second function is an aggregate, also check all the
aggregate aliases found in the first pass of the function. I'm aware that this
will not work if aggregates are nested multiple levels down, but this will have
to do for now.  It also fixes a bug where if there is a non-aggregate function
on a column in the groupby, only the original column needs to be in the groupby,
 not the entire expression. E.g. `SELECT count(), toString(stuff) BY stuff`

### DIFF
--- a/snuba_sdk/function.py
+++ b/snuba_sdk/function.py
@@ -11,7 +11,6 @@ from snuba_sdk.expressions import (
     ScalarLiteralType,
     ScalarType,
 )
-from snuba_sdk.snuba import is_aggregation_function
 
 
 class InvalidFunction(InvalidExpression):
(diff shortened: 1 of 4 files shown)


fpacifici (commented):
This seems to me another data point that hints that aliases are bad. I think this is more evidence that we should not have them at all in the sdk (as per https://github.com/getsentry/snuba-sdk/pull/41#discussion_r612000567). The only way to get them right is to port in the sdk all the complexity we have in snuba to manage them: alias expansion and mangling. I would discourage bringing that level of complexity in the sdk as we  cannot remove it from snuba. 

In alternative, in order to maintain aliases without expanding them, we would basically have to give up on a lot of the query processing and validation we would do in the sdk.

Is this still a discussion point ?

evanh:
Yeah it is still a discussion. Technically, the SDK doesn't support aliases at all right now. All of this alias code here is to support how legacy queries are using them since I need those to work without product changes. Until we have the data model validation in the SDK we won't be able to actually stop anyone from passing an alias as a column.

I am still not convinced supporting aliases is a bad thing. The SDK would be able to expand the alias without too much effort so what it sent to Snuba would have the full expression instead of the alias. Either way the point is moot until we are migrated off legacy queries, since they require alias support, and we would probably need data model validation as well so we can make sure that columns being passed in are actually columns and not aliases.

fpacifici (approved):
Ok, let's ensure there is a proper design discussion when we will have to decide whether to support aliases in the sdk or not.

PR #56, opened 2021-05-18
feat: Support embedding Expressions in tuples and arrays

evanh:
It should be possible to embed expressions into arrays, such as in the test case
added. Currently it's not possible to check the return types of the functions to
ensure that the resulting values will be correct.


snuba_sdk/expressions.py
```diff
@@ -55,11 +54,63 @@ def is_literal(value: Any) -> bool:
...
+        return f"list({find_base(to_check)})"
+
+    # Find the first non-null type
+    base_type = None
+    for elem in pot_array:
+        base_type = find_base(elem)
+        if base_type is not None:
+            break
+
+    print("BASE", base_type)
```
fpacifici, line 94: please remove

snuba_sdk/expressions.py
```diff
@@ -55,11 +54,63 @@ def is_literal(value: Any) -> bool:
...
+    if base_type is None:
+        return True
+
+    for elem in pot_array:
+        elem_type = find_base(elem)
+        print("NEXT", elem_type, elem)
```
fpacifici, line 100: remove

snuba_sdk/expressions.py
```diff
@@ -55,11 +54,63 @@ def is_literal(value: Any) -> bool:
     return isinstance(value, tuple(Scalar))
 
 
+def check_array_type(pot_array: List[Any]) -> bool:
```
fpacifici, line 57: I would advise against trying to validate the type of the elements in an array if we cannot validate all of them. I think it may be more confusing to user than actually not validating as they may expect the array to be validated while it is only in some cases.

evanh: I thought it would be worth validating the simple case still, if they mix literal types. I think the more we validate in the SDK the better, since we avoid doing a clickhouse query to discover the types are wrong.

fpacifici: I think consistency has a value as well. But I could buy this type of validation.
Actually I think there is a bigger issue I missed in my previous review:

> I think the more we validate in the SDK the better.

I am not 100% in agreement with this.
- I think it is paramount that Snuba should have the responsibility to validate everything it can.  So if we have an array type validation that snuba could implement on its own, but it is implemented only in the sdk, this seems a smell.  The issue is that while sentry is the only client today, still snuba provides an external api that can be invoked without the sdk. Thus validation should be built in Snuba first. So it would be immediately taken into account by any type of client. 
This does not mean that there should be no validation in the sdk. It makes sense to add simple validations there as well, and, more importantly, validations Snuba has no context on and thus cannot do.
- Keeping the sdk simple has a value. Imagine tomorrow Armin builds a new rust service that needs to talk to Snuba (or Chartcuterie v2 which is a Node service). We will need a Rust sdk or a javascript sdk. If we keep adding query processing logic in the sdk we would have to port all of that making the effort much harder or risking that the two sdks would diverge. If we instead we build in Snuba first, then all sdk inherit the same functionalities. 

Basically what I want to get at, is that sdk validation should be a perk (with the exception of the validation that only the sdk can do), and we should build it in Snuba first.
So why don't you write a Snuba validator with this logic instead?

evanh: While that all makes sense, I'm not sure what validation the SDK does that Snuba can't do. I guess I'd like to have a clearer understanding of what validation we should leave in vs. take out. I've tried to lean towards the idea that the SDK will generate a syntactically correct SnQL query, but not necessarily a semantically correct one. Initially I had thought having correct types in an array was a syntax error, but I can see that with the addition of Expressions it becomes a semantic (or I guess runtime) query error.

fpacifici: I was thinking about anything that snuba has no context of. 
Generally these are more product related and specific to some referrers. But you are right, this would not be standard validation provided by the sdk to all the clients, and would mostly implemented as product code by running a custom visitor on the query generated by the sdk.

Anyway, the argument I am trying to make is that, for as long as a validation is implemented in snuba first, we are good. Then I have no problem with it being in the sdk as well, as we could safely port the sdk to another language and leave that validation out since snuba does it anyway.
I would still like to keep the sdk as simple as possible, but that becomes a much smaller concern for what concerns validator  if the validation is also done by snuba and guaranteed to happen to everybody using snuba.

evanh: OK I'll remove this check for now.

fpacifici (approved):

PR #59, opened 2021-05-21
fix(legacy) Strip backticks out of aliases

evanh:
Some older discover queries were using backticks around aliases, presumably from
before the AST could properly escape aliases. They're not needed anymore.

### DIFF
--- a/snuba_sdk/legacy.py
+++ b/snuba_sdk/legacy.py
@@ -105,6 +105,8 @@ def parse_exp(value: Any) -> Any:
         return parse_scalar(value)
 
     alias = value[2] if len(value) > 2 else None
+    if alias and alias.startswith("`") and alias.endswith("`"):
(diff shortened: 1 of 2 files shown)


snuba_sdk/legacy.py
```diff
@@ -105,6 +105,8 @@ def parse_exp(value: Any) -> Any:
         return parse_scalar(value)
 
     alias = value[2] if len(value) > 2 else None
+    if alias and alias.startswith("`") and alias.endswith("`"):
+        alias = alias[1:-1]
```
fpacifici, line 109: IS it harder to actually fix this at the source ?

evanh: I don't know about harder. It was certainly faster to fix it here, rather than hunting through all of Sentry to try and find every use case of this.

fpacifici (approved):

PR #61, opened 2021-05-31
feat(subscriptions) Add flag used for legacy subscriptions

evanh:
Subscriptions should not be created with time conditions, so add a flag to the
legacy parser that skips time conditions for when Sentry uses the SDK to convert
subscriptions.

### DIFF
--- a/snuba_sdk/legacy.py
+++ b/snuba_sdk/legacy.py
@@ -173,7 +173,9 @@ def parse_condition(cond: Sequence[Any]) -> Condition:
     return Condition(parse_exp(cond[0]), Op(cond[1]), rhs)
 
 
-def json_to_snql(body: Mapping[str, Any], entity: str) -> Query:
+def json_to_snql(
+    body: Mapping[str, Any], entity: str, skip_time: bool = False
+) -> Query:
     """
     This will output a Query object that matches the Legacy query body that was passed in.
     The entity is necessary since the SnQL API requires an explicit entity. This doesn't
@@ -233,7 +235,7 @@ def json_to_snql(body: Mapping[str, Any], entity: str) -> Query:
 
     assert isinstance(query.match, Entity)
     time_column = get_required_time_column(query.match.name)
-    if time_column:
+    if time_column and not skip_time:
         time_cols = (("from_date", Op.GTE), ("to_date", Op.LT))
         for col, op in time_cols:
             date_val = body.get(col)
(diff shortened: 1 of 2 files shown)


snuba_sdk/legacy.py
```diff
@@ -242,8 +244,9 @@ def json_to_snql(body: Mapping[str, Any], entity: str) -> Query:
                     Condition(Column(time_column), op, parse_datetime(date_val))
                 )
 
-    if body.get("project"):
-        proj_cond = parse_extension_condition("project_id", body["project"], True)
+    project_id = body.get("project") or body.get("project_id")
+    if project_id:
+        proj_cond = parse_extension_condition("project_id", project_id, True)
```
fpacifici, line 249: Why this change ?

snuba_sdk/legacy.py
```diff
@@ -233,7 +235,7 @@ def json_to_snql(body: Mapping[str, Any], entity: str) -> Query:
 
     assert isinstance(query.match, Entity)
     time_column = get_required_time_column(query.match.name)
-    if time_column:
+    if time_column and not skip_time:
```
fpacifici, line 238: I think we should reject a query that tries to set time range conditions for subscriptions. Not silently drop those conditions.

tests/test_legacy_subscriptions.py
```diff
@@ -0,0 +1,99 @@
...
+            (
+                "WHERE project_id IN tuple(2) "
+                "AND type = 'error' "
+                "AND ifNull(tags[level], '') = 'error' "
+                "AND environment = 'development'"
+            ),
+        ),
+        "events",
+        id="basic_subscription_with_extra_conditions",
+    ),
```
fpacifici, line 88: Am I missing something or there is no test that is actually asserting that time conditions in the legacy query are skipped ?

snuba_sdk/legacy.py
evanh: Subscriptions use `project_id` as a key instead of `project`.

snuba_sdk/legacy.py
evanh: This isn't dropping anything. It is preventing the SDK from automatically adding them.

tests/test_legacy_subscriptions.py
evanh: The fact that the time conditions are not in the generated SnQL query is the test.

snuba_sdk/legacy.py
fpacifici: Can't the client decide not to add `from_date` or `to_date` to the query body?
The loop below only adds the conditions if those fields are in the query body in the first place.
If no `from_date` and `to_date` are present it would not add anything, so `if skip_time`, we would ignore those fields.

Shouldn't we reject the query if the user provided `from_date` and `to_date` in the legacy query ?

evanh: Oh I see what you're saying now. You're right this change isn't needed. If someone tries to add a time condition to a subscription, then the validation in Snuba will catch it and error. So the SDK can work as normal.

evanh:
I'm going to close this PR because I don't think it's necessary. The `skip_time` field isn't very helpful since the conditions are already skipped if `from_date` and `to_date` are missing from the body, so the SDK will work as expected in that case. I will change Sentry to pass the project ID in the `project` field to fix that issue.

PR #62, opened 2021-06-04
fix(legacy) Handle weird edge case with escaping

evanh:
There seems to be a bug in Sentry that sends search strings like this, but since
the legacy endpoint would allow this the SDK should make it work. Sometimes a
string is received with a single backslash on the end, which was incorrectly
escaping the single quote around the string and causing a parsing error. If
there is a trailing backslash, escape it.

### DIFF
--- a/tests/test_legacy_api.py
+++ b/tests/test_legacy_api.py
@@ -1183,6 +1183,51 @@ discover_tests = [
         "events",
         id="dots_in_alias",
     ),
+    pytest.param(
+        {
+            "project": [1],
+            "dataset": "events",
+            "from_date": "2021-04-01T20:05:27",
+            "to_date": "2021-04-15T20:05:27",
+            "groupby": ["group_id"],
+            "conditions": [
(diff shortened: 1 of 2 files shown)


tests/test_legacy_api.py
```diff
@@ -1183,6 +1183,51 @@
...
     ),
+    pytest.param(
+        {
+            "project": [1],
+            "dataset": "events",
+            "from_date": "2021-04-01T20:05:27",
+            "to_date": "2021-04-15T20:05:27",
+            "groupby": ["group_id"],
+            "conditions": [
+                [["positionCaseInsensitive", ["message", "'Api\\'"]], "!=", 0],
```
evanh, line 1194: I think this is a Sentry bug by not escaping the trailing backslash, but it's easier to fix it here then try to dig into the internals of issues and group serialization. It's a VERY rare edge case.

fpacifici (approved):

PR #66, opened 2021-06-15
feat: Allow importing from snuba_sdk directly

evanh:
This is a big time saver for users of the SDK since it cleans the code up and
means that users don't have to hunt down which package certain classes are in.
Some things that are meant for special cases (DryRun, Legacy) are not added.

### DIFF
--- a/snuba_sdk/query.py
+++ b/snuba_sdk/query.py
@@ -1,26 +1,24 @@
 from dataclasses import dataclass, fields, replace
 from typing import Any, List, Optional, Sequence, Union
 
-from snuba_sdk import (
-    BooleanCondition,
-    Column,
-    Condition,
+from snuba_sdk.column import Column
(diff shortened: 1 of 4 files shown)


lynnagara (approved):

snuba_sdk/query.py
```diff
@@ -1,24 +1,26 @@
 from dataclasses import dataclass, fields, replace
 from typing import Any, List, Optional, Sequence, Union
 
-from snuba_sdk.column import Column
-from snuba_sdk.conditions import BooleanCondition, Condition
-from snuba_sdk.entity import Entity
-from snuba_sdk.expressions import (
```
lynnagara, line 7: Probably need to keep the original imports in this file

evanh: Yeah I meant to do this in a test file, not here.

PR #67, opened 2021-06-16
fix(legacy) Allow quotes in functions

evanh:


fpacifici (approved):
I think we should revert this special case when we stop using the legacy representation

PR #68, opened 2021-06-23
fix(escaping) Properly escape strings

evanh:
The visitor now correctly escapes ' \ and \n. Snuba requires a change to correctly handle these values as well, so once this is in a release, we'll have to coordinate using it in Sentry and Snuba.

### DIFF
--- a/snuba_sdk/visitors.py
+++ b/snuba_sdk/visitors.py
@@ -1,4 +1,3 @@
-import re
 from abc import ABC, abstractmethod
 from datetime import date, datetime
 from typing import Any, Generic, Set, TypeVar
@@ -25,11 +24,6 @@ from snuba_sdk.expressions import (
 from snuba_sdk.function import CurriedFunction, Function
 from snuba_sdk.orderby import LimitBy, OrderBy
 from snuba_sdk.relationships import Join, Relationship
-
-# validation regexes
-unescaped_quotes = re.compile(r"(?<!\\)'")
-unescaped_newline = re.compile(r"(?<!\\)\n")
-
 
 TVisited = TypeVar("TVisited")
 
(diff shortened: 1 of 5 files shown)


snuba_sdk/visitors.py
```diff
@@ -157,8 +151,12 @@ def _stringify_scalar(self, value: ScalarType) -> str:
             else:
                 decoded = value
 
-            decoded = unescaped_quotes.sub("\\'", decoded)
-            decoded = unescaped_newline.sub("\\\\n", decoded)
+            # The ' and \ character are escaped in the string to ensure
+            # the query is valid. They are de-escaped in the SnQL parser.
+            # Also escape newlines since they break the SnQL grammar.
+            decoded = (
```
fpacifici, line 157: You can replace multiple characters with a single pass with `maketrans` 
https://www.w3schools.com/python/ref_string_translate.asp

tests/test_scalar.py
```diff
@@ -33,7 +33,7 @@
     pytest.param("abc", "'abc'"),
     pytest.param(b"abc", "'abc'"),
     pytest.param("a'b''c'", "'a\\'b\\'\\'c\\''"),
-    pytest.param("a\\''b''c'", "'a\\'\\'b\\'\\'c\\''"),
+    pytest.param("a\\''b''c'", "'a\\\\\\'\\'b\\'\\'c\\''"),
```
fpacifici, line 36: Maybe I am missing something obvious but why do we escape special characters with two backslashes instead of one when generating SnQL ?
Like can't we escape 
```
`
```
as 
```
\`
```
?

evanh: That's what we do. In Python I have to escape the backslash again.

fpacifici (approved):

PR #70, opened 2021-07-05
fix(validate) Remove groupby checks

evanh:
This code has been brittle and difficult to follow for a while, but what pushed
this change over the edge was queries being written for sessions. The `sessions`
column in Sentry is an alias for an aggregate function in Snuba, but the SDK
doesn't know that so it treats it as a non-aggregate and gets upset if it's not
included in the group by. So the SDK would need to know about all such aliases
in Snuba and it's just not worth it. Let Snuba handle this validation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changelog and versioning
 ==========================
 
+0.0.20
+------
+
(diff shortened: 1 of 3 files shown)


iProgramStuff (approved):
Awesome! Thanks Evan.

fpacifici (approved):

PR #72, opened 2021-07-20
fix(legacy) Be less restrictive on aliases

evanh:
This came up because Sentry was using `equation[0]` as an alias. Ultimately, since we escape aliases with backticks in Snuba, it didn't seem like I should block any alias, except one that would break our own aliases (hence why I blocked the character set ``[`'"\]``

### DIFF
--- a/snuba_sdk/function.py
+++ b/snuba_sdk/function.py
@@ -17,7 +17,7 @@ class InvalidFunction(InvalidExpression):
     pass
 
 
-alias_re = re.compile(r"^[a-zA-Z]([^\s`'\"])*$")
+alias_re = re.compile(r"^[a-zA-Z]([^\s`'\\\"])*$")
 # In theory the function matcher should be the same as the column one.
 # However legacy API sends curried functions as raw strings, and it
 # wasn't worth it to import an entire parsing grammar into the SDK
--- a/snuba_sdk/function.py
+++ b/snuba_sdk/function.py
@@ -17,7 +17,7 @@ class InvalidFunction(InvalidExpression):
     pass
 
 
-alias_re = re.compile(r"^[a-zA-Z](\w|\.)+$")
+alias_re = re.compile(r"^[a-zA-Z]([^\s`'\"])*$")
 # In theory the function matcher should be the same as the column one.
 # However legacy API sends curried functions as raw strings, and it
 # wasn't worth it to import an entire parsing grammar into the SDK
--- a/snuba_sdk/function.py
+++ b/snuba_sdk/function.py
@@ -17,7 +17,7 @@ class InvalidFunction(InvalidExpression):
     pass
 
 
-alias_re = re.compile(r"^[a-zA-Z]([^\s`'\\\"])*$")
+alias_re = re.compile(r"^[a-zA-Z](\w|\.|\[|\])+$")
 # In theory the function matcher should be the same as the column one.
 # However legacy API sends curried functions as raw strings, and it
 # wasn't worth it to import an entire parsing grammar into the SDK
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ Changelog and versioning
 0.0.22
 ------
 
-- Make aliases less restrictive
+- Allow square brackets in aliases
 
 0.0.21
 ------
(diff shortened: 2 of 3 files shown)


vladkluev:
Can you write in the PR what was possible before this PR and what is possible now?

snuba_sdk/function.py
```diff
@@ -17,7 +17,7 @@ class InvalidFunction(InvalidExpression):
     pass
 
 
-alias_re = re.compile(r"^[a-zA-Z](\w|\.)+$")
+alias_re = re.compile(r"^[a-zA-Z]([^\s`'\"])*$")
```
fpacifici, line 20: Which characters was the product using that we were not allowing?
This seem very permissive now as you can declare an alias `f(x)` when `f(x)` is a valid expression otherwise you probably do not want to shadow with an alias

lynnagara:
Is this based on a list that clickhouse defines somewhere?

snuba_sdk/function.py
evanh: I updated the PR comment. This is inline with what we allowed as aliases in the legacy syntax, since we escape aliases with backticks.

evanh:
> Is this disallowed characters based on a list that clickhouse defines somewhere?

Partially. We escape aliases with backticks in Snuba, so if an alias already has a backtick that should not be allowed. However the other characters are special characters in SnQL so I disallowed them as well.

snuba_sdk/function.py
fpacifici: Right but in the old syntax the structure of the query itself (it was json) defined the tokens so, except when the users abused the aggregation field, you could not have a conflict between `f(x)` function call and `f(x)` alias as the function call had a json structure.

I don't think we should stick to that behavior for the sake of allowing a behavior that was ok in the old language but would become a smell in this language.

Can we fix the alias names in the product when they use square brackets or parentheses ?

fpacifici: Are these queries generated by the automatic translation from legacy to SnQL ?

lynnagara:
I'm curious why spaces are not allowed as it seems we are attempting to be as permissive as possible. As an example in Discover `select event_id as `Event ID` seems like it would be pretty useful.

snuba_sdk/function.py
evanh: I don't know why  the automatic translation vs. use of SnQL matters, but they are using the SDK directly.

Clickhouse has no limitations on what can be an alias, as long as you escape the alias with backticks. Snuba similarly has no restrictions since we escape backticks and allow any other characters.

https://github.com/getsentry/snuba/blob/master/snuba/clickhouse/escaping.py#L7

I'm not sure what you suggest I do, if someone wants to have an alias of `f(x)` then that should be allowed by us.

evanh:
Snuba actually does escape the backticks, so that's not an issue. You're right that it would be better, but that would require changes to the SnQL grammar as well and I'm trying to get this fix out quickly to fix a prod issue.

lynnagara (approved):
This seems good to me, though I think we should eventually add the handling for spaces, backticks and backslashes too.

lynnagara:
Actually, one other question - why does an alias have to start with a character in the range [a-zA-Z]?

snuba_sdk/function.py
fpacifici: > I don't know why the automatic translation vs. use of SnQL matters,
It matters because if they are using snql directly they can just change their alias right away, if it was an automatic translation, by preventing this, we would have made a valid query (in the old system) that had no reason to be rejected invalid.

> if someone wants to have an alias of f(x) then that should be allowed by us.
That is what I disagree on. I think it is a bad practice we can prevent.

fpacifici: Also I think it may not be trivial to tell alias references (in the case of `f(x)` or `tags[asd]` used as aliases) apart from a function call or a tag reference during parsing unless we require them to be escaped.

evanh:
> Actually, one other question - why does an alias have to start with a character in the range [a-zA-Z]?

I think I took that from Clickhouse, but reading it over again that rule only applies to aliases that aren't wrapped in backticks. This needs to all be revisited and cleaned up.

PR #73, opened 2021-07-23
fix(legacy) Fix legacy nested functions

evanh:
In very rare cases (only Sentry tests as far as I can tell) legacy functions are
nested ambiguously, without enclosing brackets to delineate the nested function.
The legacy parser had a special case to handle this. Import that functionality
to the legacy parser and use it when the original function fails.

### DIFF
--- a/snuba_sdk/legacy.py
+++ b/snuba_sdk/legacy.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from functools import partial
-from typing import Any, List, Mapping, Optional, Sequence, Union
+from typing import Any, List, Mapping, Optional, Sequence, Tuple, Union
 
 from snuba_sdk.column import Column
 from snuba_sdk.conditions import Condition, Op, Or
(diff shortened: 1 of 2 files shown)


tests/test_legacy_api.py
```diff
@@ -1367,6 +1367,49 @@
...
+                [
+                    "if",
+                    [
+                        ["in", ["tags[sentry:release]", "tuple", ["'foo'"]]],
+                        "tags[sentry:release]",
+                        "'other'",
+                    ],
+                    "release",
+                ]
+            ],
```
fpacifici, line 1382: Could you provide an example of how this was parsed before ? I am not sure about about what the specific issue was. I remember ambiguities in the parsing, but cannot remember the details.

snuba_sdk/legacy.py
```diff
@@ -107,23 +184,17 @@ def parse_exp(value: Any) -> Any:
...
-        children = [parse_exp(value[1])]
+def parse(value: Any) -> Any:
+    if not isinstance(value, list):
+        return parse_exp(value)
 
-    return Function(value[0], children, alias)
+    try:
+        return parse_exp(value)
+    except Exception:
+        return parse_legacy_function(value)
```
fpacifici, line 197: Which exception do you expect here ?

tests/test_legacy_api.py
evanh: The original parser would fail because it would recurse through the first term until it got to `"in", ["tags[sentry:release]", "tuple", ["'foo'"]]]`, where it would call `parse_exp` on each of the terms individually. This would throw an exception because `["'foo'"]` is a list, which it would try to interpret as a function and fail.

In Snuba, the parser iterates through the arguments by one but also by two, and evaluate whether those two arguments are in turn a function.

snuba_sdk/legacy.py
evanh: The specific error I was dealing with threw an IndexError, but I figured that really any error should be cause to try the legacy parser.

fpacifici: Why don't we only run the legacy parser? 
If the old language behaved in that ambiguous way and this is meant to translate legacy queries into SnQL queries, I think we should just run the old parsing logic here. 
Relying on an exception to assume the issue is that we are using the wrong parser seems a very fragile and ad-hoc solution that would shadow real issues that manifest themselves in the same way by raising an Exception.
It also seems a smell that the parser relies on a function to raise an exception to do its normal behavior.

If the production environment depends on the non ambiguous syntax to translate legacy queries to SnQL query it seems to me we are doing something wrong.

evanh: I agree this solution isn't ideal. I originally had the same idea you did, which was to solely use the logic from the legacy Snuba parser. However a large number of my tests started failing when I did that, with the resulting SnQL being inconsistent with the previous parsing. I am worried that if I change all this parsing now, I'm going to introduce bugs that weren't there before and will be harder to track since I removed the dry run SQL comparison.

evanh: The more I think about this, the more I realize this: This change is purely to support a single Sentry test. It has not occurred in production yet. I'm going to try to fix the test to be unambiguous and see if this is actually something that could occur in production or is unique to the test.

evanh:
I'm going to close this, I changed the test in Sentry to be properly formatted. I think the Sentry test was just poorly written and isn't a case that occurs in production.

PR #75, opened 2021-07-29
feat: Add an output_alias to Column

evanh:
This is used to add an "AS x" clause to Columns. The output_alias is only used
when translating the select and groupby expressions, so the alias should never
appear in other clauses.


snuba_sdk/column.py
```diff
@@ -32,6 +32,7 @@ class Column(Expression):
 
     name: str
     entity: Optional[Entity] = None
+    output_alias: Optional[str] = None
```
vladkluev, line 35: I would leave a comment here explaining what an output alias is as it is not very obvious

vladkluev: Also, how should it show up in the string representation of the expression?

snuba_sdk/expressions.py
```diff
@@ -17,6 +18,8 @@ def validate(self) -> None:
         raise NotImplementedError
 
 
+alias_re = re.compile(r"^[a-zA-Z](\w|\.|\[|\])*$")
```
vladkluev, line 21: I've used the convention before that read only globals are in all caps. Is that not the convention here?

tests/test_query_visitors.py
```diff
@@ -350,6 +350,54 @@
...
+        .set_groupby(
+            [
+                Column("project_id", output_alias="pi"),
+                Column("transaction", output_alias="tn"),
+            ]
+        )
+        .set_where([Condition(Column("project_id"), Op.IN, (1,))]),
+        (
+            "MATCH (events)",
+            "SELECT transaction AS tn, count() AS equation[0]",
```
vladkluev, line 370: The column has an alias which is not the same as an output alias but it shows up the same way in the query?

vladkluev (commented):
I don't have any issues with the code but the concept of the output alias is not clear to me

CHANGES.rst
```diff
@@ -4,7 +4,8 @@ Changelog and versioning
 0.0.23
 ------
 
-- Fix bug with legacy queries sending nested functions in a different structure
```
fpacifici, line 7: Why did you delete this line ?

tests/test_query_visitors.py
evanh: Columns do not have aliases in the SDK. That's what this PR is (partially) adding.

snuba_sdk/column.py
fpacifici: You can use columns everywhere in the query. While the output_alias would only be allowed for columns that are in the top level select clause and top level group by. This makes me think that the place for such alias is not a field in the column but another data structure like we do for the query object in snuba.

Why don't we change the type of the select clause and the group by clause from a sequence of Expresison to a Serquence of AliasedExpressions (equivalent of SelectedExpression in the snuba AST). Where AliasedExpression contains an expression and provides the alias. 

That is supposed to make the usage of the Column class structurally sound (you cannot structurally add an alias in clauses where that is forbidden without having to add additional validation at runtime). And you do not need to manage the output alias separately in all types of expressions you can have in the select clause.

evanh: I added a comment about what an output alias is for. It shows up in the query as `name AS alias`.

snuba_sdk/visitors.py
```diff
@@ -182,10 +184,15 @@ def _stringify_scalar(self, value: ScalarType) -> str:
...
-        alias_clause = ""
+        entity_alias_clause = ""
         if column.entity is not None and self.use_entity_aliases:
-            alias_clause = f"{column.entity.alias}."
-        return f"{alias_clause}{column.name}"
+            entity_alias_clause = f"{column.entity.alias}."
+
+        alias_clause = ""
+        if column.output_alias and self.use_output_aliases:
+            alias_clause = f" AS {column.output_alias}"
```
fpacifici, line 193: Having to rely on state changes of the translator to decide if to apply the alias is also something you would not need to do if we had a specific type for AliasedExpressions

fpacifici (changes requested):
I don't think the alias should be a field of the Column class but a wrapper of the expression you want to alias (like we do in the AST for SelectedExpresison). See my comments inline for more details.

CHANGES.rst
evanh: That was the PR I closed because I fixed it on Sentry's side. https://github.com/getsentry/snuba-sdk/pull/73

snuba_sdk/column.py
evanh: The reason I did things this way was to make it easier to reuse Expressions throughout queries. Having this field be in the Column class but optional allows users to reuse the Column everywhere in the Query object without having to redefine it or wrap it in a class.

I'll try with the AliasedExpression and see if that works in a better way.

snuba_sdk/aliased_expression.py
```diff
@@ -0,0 +1,37 @@
...
+    anywhere in Snuba except to change the names in the results set. Right now this is
+    limited to Columns only because Functions have a separate alias. Eventually the
+    two will be combined.
+
+    :param Expression: The expression to alias.
+    :type Expression: Column
+    :raises InvalidExpression: If the expression or alias is invalid.
+    """
+
+    # TODO:
```
fpacifici, line 21: TODO ?

fpacifici (approved):

PR #78, opened 2021-09-02
feat(parent_api) Support the parent_api parameter

evanh:
Support passing this parameter into queries, legacy and otherwise.

Related to https://github.com/getsentry/snuba/pull/2076


vladkluev (approved):
LGTM

PR #82, opened 2021-11-11
ref(3.9) Upgrade snuba-sdk to use python 3.9

evanh:
The sentry repository has been updated to 3.8, so we can now finally update and
drop support for 3.6. This also means we can use annotations which is nice.


vladkluev:
If sentry was updated to 3.8, why are we upgrading to 3.9?

evanh:
I was using 3.9 since we can, but in hindsight we should probably default to what our other packages default too.

PR #85, opened 2021-11-24
Feat: Add required columns to Entity data models

evanh:
The EntityModel now can have required columns and a required time column. These
columns must be present as conditions in the query. They are validated at the
same time that the match clause is validated.


snuba_sdk/query_validation.py
```diff
@@ -86,15 +100,123 @@ def _validate_join(match: Join, all_columns: Set[Expression]) -> None:
...
+
+class RequiredColumnError(Exception):
+    pass
+
+
+def _check_entity_required_columns_in_conditions(
+    entity: Entity, conditions: Optional[Sequence[Union[BooleanCondition, Condition]]]
+) -> None:
+    """
+    For a given entity, flatten the AND conditions to find just the top level ones and
```
vladkluev, line 127: The `validate_required_columns` docstring gives almost all the context that this function does. It may be a good idea to consolidate the two under the public module function and just point to it from here. The example is good though so you should definitely keep that

snuba_sdk/schema.py
```diff
@@ -15,6 +16,9 @@ class ColumnSet:
     def __init__(self, columns: Sequence[Column]) -> None:
         super().__setattr__("column_names", set(c.name for c in columns))
         super().__setattr__("columns", set(columns))
+        super().__setattr__(
+            "required_columns", set([r for r in self.columns if r.required])
+        )
```
vladkluev, line 21: Unrelated to this PR but I think it's worth raising. 

We have all this fancy code to set stuff on the class but we wouldn't need it if the dataclass was not frozen. Why do we mark it as frozen and then mutate it?

vladkluev (approved):
Thanks for splitting this up. I think you have some artifacts from doing it but those should be easy to clean up. Otherwise this looks great.

PR #86, opened 2021-11-25
ref(typing) Add a type for sequences of conditions

evanh:
Create a type alias to use for the sequence of conditions a query can have in
the where or having clause. This has no functional changes, it just reduces the
verbosity of the typing.


snuba_sdk/conditions.py
```diff
@@ -133,17 +133,16 @@ def validate(self) -> None:
...
 @dataclass(frozen=True)
 class Or(BooleanCondition):
     op: BooleanOp = field(init=False, default=BooleanOp.OR)
-    conditions: Sequence[Union[BooleanCondition, Condition]] = field(
-        default_factory=list
-    )
+    conditions: ConditionSet = field(default_factory=list)
+
+
+ConditionSet = Sequence[Union[BooleanCondition, Condition]]
```
vladkluev, line 145: ConditionSet implies uniqueness but this is not a `set`, it's a sequence
```suggestion
ConditionSequence = Sequence[Union[BooleanCondition, Condition]]
```

vladkluev: As a side note, I do find it strange that `BooleanCondition`s and `Condition`s are different things but that's out of scope for this PR

vladkluev (changes requested):
I appreciate the cleanup here a lot but I don't think the name is quite right

PR #89, opened 2021-11-30
fix(precommit) Fix the black check for precommit on M1

evanh:
Unfortunately I haven't figured out why mypy fails in precommit. I might have to wait for the next mypy release.


vladkluev (approved):
thanks! Just one small thing you forgot while testing

PR #91, opened 2022-01-12
feat: Add Lambdas and Identifiers expressions

evanh:
Add Lambda and Identifier expressions to the syntax. Lambdas can be used to
build anonymous functions in the form (`x`) -> <transformation>. Identifiers
can be used inside the transformation to reference the identifier(s) in the
lambda. These can be used inside functions to support higher order functions
such as arrayMap.

________

```
Function(
    "arrayMap", 
    Lambda(
        ["x"], 
        Function(
            "toString", 
            [Identifier("x")]
        )
    ), 
    Column("some_array")
)
```

### DIFF
--- a/snuba_sdk/function.py
+++ b/snuba_sdk/function.py
@@ -33,7 +33,9 @@ class CurriedFunction(Expression):
     function: str
     initializers: Optional[Sequence[Union[ScalarLiteralType, Column]]] = None
     parameters: Optional[
-        Sequence[Union[ScalarType, Column, CurriedFunction, Function]]
+        Sequence[
+            Union[ScalarType, Column, CurriedFunction, Function, Identifier, Lambda]
+        ]
     ] = None
     alias: Optional[str] = None
 
@@ -79,7 +81,7 @@ class CurriedFunction(Expression):
                 )
             for param in self.parameters:
                 if not isinstance(
-                    param, (Column, CurriedFunction, Function)
+                    param, (Column, CurriedFunction, Function, Identifier, Lambda)
                 ) and not is_scalar(param):
                     assert not isinstance(param, bytes)  # mypy
                     raise InvalidFunctionError(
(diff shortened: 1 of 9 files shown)


MeredithAnya:
@evanh you use `arrayMax` in the PR description, and `arrayMap` in the test functions, did you mean just `arrayMap`? Do those function(s) need to be added to the global allow list?

snuba_sdk/function.py
```diff
@@ -103,3 +105,34 @@ class Function(CurriedFunction):
...
+    name: str
+
+    def validate(self) -> None:
+        if not isinstance(self.name, str) or not identifier_re.match(self.name):
+            raise InvalidLambdaError(f"`{self.name}` is not a valid identifier")
+
+
+@dataclass(frozen=True)
+class Lambda(Expression):
+    identifiers: Sequence[str]
```
nikhars, line 128: Should this be a sequence of Identifier which you defined above?

snuba_sdk/function.py
```diff
@@ -103,3 +105,34 @@ class Function(CurriedFunction):
...
+    transformation: CurriedFunction
+
+    def validate(self) -> None:
+        if not isinstance(self.identifiers, (tuple, list)):
+            raise InvalidLambdaError("identifiers must be a sequence")
+        for i in self.identifiers:
```
nikhars, line 134: If self.identifiers is going to be `Identifier` then you can just call `self.identifiers.validate` here.

evanh:
> @evanh you use `arrayMax` in the PR description (the text not the example), and `arrayMap` in the test functions, did you mean just `arrayMap`? Do those function(s) need to be added to the global allow list?

They are both valid examples, but I changed the description to be consistent.

As for adding to the allowed functions, I didn't think of that. I'll add them.

snuba_sdk/function.py
evanh: No, I deliberately left it a string so that the construction of a Lambda looks like `Lambda(["x", "y"], ...)` instead of `Lambda([Identifier("x"), Identifier("y")], ...)`. It's unnecessary to define the type there since they can't be anything but identifiers.

snuba_sdk/function.py
evanh: I would like to leave them as strings (see above) but also I couldn't call `self.identifiers.validate` directly since that property is a Sequence which doesn't have a validate function.

nikhars (approved):

PR #92, opened 2022-01-27
feat(attribution) Add team and feature tag to API

evanh:
The team and feature tag will be passed in from Sentry and used in Snuba as
attribution metrics. Add them to the API so Sentry can start populating them.

Also do some refactoring of flags in general, to help make it easier to add
different flags in the future.


snuba_sdk/query_visitors.py
```diff
@@ -204,6 +211,8 @@ def _combine(self, query: main.Query, returns: Mapping[str, str]) -> str:
             "dry_run",
             "legacy",
             "parent_api",
+            "team",
```
evanh, line 214: This check is a legit gotcha if you aren't aware of it. I think the way the Query stores flags needs to be refactored to make the logic easier to understand.

snuba_sdk/query_visitors.py
```diff
@@ -335,7 +350,11 @@ def _combine(self, query: main.Query, returns: Mapping[str, str]) -> str:
         if query.legacy:
             body["legacy"] = query.legacy.value
         if query.parent_api:
-            body["parent_api"] = query.parent_api.name
+            body["parent_api"] = query.parent_api.value
+        if query.team:
+            body["team"] = query.team.value
+        if query.feature:
+            body["feature"] = query.feature.value
```
vladkluev, line 357: Something that bothers me about this:

Why are these flags expressions when they're not part of the query? They're attached to the query but then we move it outside the query when we serialize it. Like theoretically you can have a subquery with flags even though that won't be honored (since it doesn't make sense). Why is it done this way?

evanh: It's essentially a consequence of the legacy query and how this SDK was integrated. Since this library was added as a way to replace just the "query" part of the request, I only added one top level object, the `Query`. However it became clear as I was working that this library would also need to be able to handle the extra flags as well, particularly for converting legacy queries into SnQL.

I agree it doesn't make sense and it needs to be refactored. As we've expanded the flags this model makes less and less sense and adds to our tech debt. I think particularly if we make the SDK responsible for building the request body separately from the Query, we can better handle these flags.

vladkluev (approved):
Definitely should get the flags away from the query but not in this PR. That requires the sdk to be the snuba-driver

PR #94, opened 2022-03-04
feat(expression): Add parentheses to alias regex

ahmedetefy:
This PR:
- Adds parentheses to alias regex which is quite helpful to add as in the case of running conditional aggregate in queries over metrics, it is often required to disambiguate the columns returned depending on the aggregation, and the metric name 
- Adds backticks surrounding the alias

As an example, lets say we have a query like 
```
SELECT 
    sumMergeIf((value AS _snuba_value), equals((metric_id AS _snuba_metric_id), 150) AS sum,
    sumMergeIf((value AS _snuba_value), equals((metric_id AS _snuba_metric_id), 149)) As sum
FROM metrics_counters_local
WHERE ....
```
The most convenient way is to disambiguate these two fields in the query results is according to the aggregate and the metric name
If 150 -> session, and 149 -> user, then re-write this query as 
```
SELECT 
    sumMergeIf((value AS _snuba_value), equals((metric_id AS _snuba_metric_id), 150) AS sum(session),
    sumMergeIf((value AS _snuba_value), equals((metric_id AS _snuba_metric_id), 149)) As sum(user)
FROM metrics_counters_local
WHERE ....
```
I have spoken to @evanh about this, and he told me that if Clickhouse supports this (which it does), no reason why we shouldn't support it in our alias regex


evanh (approved):
Can you please add a new version entry to CHANGES.rst, and describe this change. Once this is merged we can release that new version.

PR #99, opened 2022-04-27
ref: Update docs with the Request class and examples

evanh:


snuba_sdk/query_visitors.py
```diff
@@ -509,29 +343,5 @@ def _visit_offset(self, offset: Optional[Offset]) -> None:
...
-    def _visit_debug(self, debug: Debug) -> None:
-        debug.validate()
-
-    def _visit_dry_run(self, dry_run: DryRun) -> None:
-        dry_run.validate()
-
-    def _visit_legacy(self, legacy: Legacy) -> None:
-        legacy.validate()
+    def _visit_totals(self, totals: Optional[Totals]) -> None:
+        totals.validate() if totals is not None else None
```
volokluev, line 347: Is totals a flag or an expression? I'm not really clear on which it is

tests/test_flags.py
```diff
@@ -1,63 +1,31 @@
...
-    if flag != ParentAPI:
-        with pytest.raises(
-            InvalidExpressionError,
-            match=re.escape(f"{name} contains invalid characters"),
-        ):
-            flag("`````")
+def test_flag_with_none() -> None:
+    f = Flags(debug=None, totals=False, dry_run=True)
+    f.validate()
+    assert f.to_dict() == {"totals": False, "dry_run": True}
```
volokluev, line 31: like over here, why is totals in the flags and in the query?

tests/test_legacy.py
```diff
@@ -298,9 +298,39 @@
...
+    assert request.flags is not None
+    assert request.flags.legacy is True
+
+    query_clauses = [c for c in clauses if not c.startswith("--")]
+    query_expected = "\n".join(query_clauses)
+    assert request.query.print() == query_expected
+
+    # Rather than rewrite all these tests to have the new format, transform the
+    # old format to the new one. This captures all the flags that were using
+    # the -- print format, and asserts they are in the request flags
```
volokluev, line 313: let's make sure we have a ticket to convert the tests rather than doing this processing

volokluev (commented):
A few questions. Also this is very similar to the previous PR, any changes in particular I should be looking at?

evanh:
@volokluev I just realized I messed up the rebasing with this PR. It should be purely docs changes.

volokluev (approved):

PR #101, opened 2022-05-05
chore: Update logo in readme

mattjohnsonpint:
Switch to using the purple Sentry logo in the readme, as the black one does not render well on a black background, which will happen if you are using a dark theme in your GitHub settings.

Since the readme also renders on PyPi, where the system theme is ignored, the purple logo will look good there as well.

Also updated the sentry url to include utm source/medium parameters, as we do on the other repos.

Thanks.

#skip-changelog


mattjohnsonpint:
See also https://github.com/getsentry/sentry/pull/34229 and https://github.com/getsentry/sentry-electron/pull/479#issuecomment-1118755203 for more detail.

onewland (approved):

PR #102, opened 2022-08-12
feat(column): Expand column regex [TET-323]

ahmedetefy:
Expands columns regex to accept "/" and "@"
chars to support MRI formats for column names
such as "d:sessions/duration@millisecond"

Counterpart to https://github.com/getsentry/snuba/pull/3045

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Changelog and versioning
 ==========================
 
+1.0.1
+------
+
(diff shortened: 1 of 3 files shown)


evanh (approved):
Can you also add a new version entry in CHANGES.md, and we can cut a new release for this change.

Note that there are two other changes tied to this: once we release the new version of the SDK, both Snuba and Sentry will need to be changed to use the new version.

PR #103, opened 2022-09-12
ref: upgrade linters

asottile-sentry:


linter-requirements.txt
```diff
@@ -1,7 +1,6 @@
-black==22.3.0
-flake8==4.0.1
-flake8-import-order==0.18.1
```
asottile-sentry, line 3: this was not enabled, plus `isort` is already fixing / enforcing this so I removed it

.pre-commit-config.yaml
```diff
@@ -27,9 +27,8 @@ repos:
       - id: isort
         name: isort (python)
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v0.812'
+    rev: 'v0.971'
     hooks:
     -   id: mypy
-        args: [--config-file, mypy.ini, --ignore-missing-imports]
```
asottile-sentry, line 33: this is the default so I removed it

.flake8
```diff
@@ -1,16 +1,7 @@
 [flake8]
-ignore =
```
asottile-sentry, line 2: extend-ignore is better than ignore since it preserves the default ignore list

.flake8
```diff
@@ -1,16 +1,7 @@
...
-  W503,  // Handled by black (Line break occured before a binary operator)
-  E402,  // Sometimes not possible due to execution order (Module level import is not at top of file)
-  E731,  // I don't care (Do not assign a lambda expression, use a def)
-  B950,  // Handled by black (Line too long by flake8-bugbear)
-  B011,  // I don't care (Do not call assert False)
-  B014,  // does not apply to Python 2 (redundant exception types by flake8-bugbear)
-  N812,  // I don't care (Lowercase imported as non-lowercase by pep8-naming)
-  N804   // is a worse version of and conflicts with B902 (first argument of a classmethod should be named cls)
-max-line-length = 80
-select = N,B,C,E,F,W,T4,B9
```
asottile-sentry, line 15: plugins are enabled by default when installed so I removed this

.flake8
```diff
@@ -1,16 +1,7 @@
...
-exclude=checkouts,lol*,.tox
+extend-ignore =
+  # Handled by black (Whitespace before ':' -- handled by black)
+  E203,
+  # Handled by black (Line too long)
+  E501,
+extend-exclude=checkouts,lol*
```
asottile-sentry, line 7: extend-exclude is better than exclude since it preserves the default exclude list

.flake8
```diff
@@ -1,16 +1,7 @@
 [flake8]
-ignore =
-  E203,  // Handled by black (Whitespace before ':' -- handled by black)
-  E266,  // Handled by black (Too many leading '#' for block comment)
-  E501,  // Handled by black (Line too long)
-  W503,  // Handled by black (Line break occured before a binary operator)
-  E402,  // Sometimes not possible due to execution order (Module level import is not at top of file)
```
asottile-sentry, line 7: ini comments are either `#` or `;` not `//` -- these are actually being interpreted as ignore codes https://flake8.pycqa.org/en/latest/user/configuration.html

evanh (approved):

PR #112, opened 2023-02-14
feat(CapMan): Add tenant_ids to Request object

rahul-kumar-saini:
### Overview
- As part of the Capacity Management project, we want to consolidate "tenant" (or general grouping of a query eg referrer, org_id, etc) information into a single dictionary in the Request
- This would be passed by Sentry to Snuba, which will use this information to manage capacity (or rate limit) in a more organized and methodical fashion
- As of now there are no constraints on what these tenants can be

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Changelog and versioning
 ==========================
 
+1.0.4
+------
+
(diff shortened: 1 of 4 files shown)


volokluev (approved):

evanh (approved):
Make sure to update CHANGES.md before you release a new version with this.

PR #113, opened 2023-02-17
fix(CapMan): `tenant_ids` field now optional to support gradual adoption

rahul-kumar-saini:
### Overview
- Making this field required right away makes it so updating the Sentry repo with a new snuba-sdk version creates a huge set of changes which would be very hard to review
- This way also allows teams to pass tenant_ids themselves on their own time

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changelog and versioning
 ==========================
 
+1.0.5
+------
+
+- `tenant_ids` required field added previously is now optional to support gradual adoption
+
 1.0.4
 ------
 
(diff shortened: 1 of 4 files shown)


volokluev (commented):
The field is optional now but can we have the sdk log the request when the tenant ids are not provided? might make things easier to transition

volokluev (approved):
approved but consider my comment

rahul-kumar-saini:
> The field is optional now but can we have the sdk log the request when the tenant ids are not provided? might make things easier to transition

The SDK doesn't really have any access to logging/metrics as far as I can tell. I do however intend to create a metric on Snuba which will track Requests with vs without tenant_ids, tagged with referrer so we can gradually have queries from different features onboard.

PR #121, opened 2023-09-19
feat(metrics) Add translators/validators for MetricsQuery

evanh:
Add the basic SnQL translators for the MetricsQuery object and all the objects
it uses.

This adds a translator and validator visitor for the MetricsQuery class. It
also changes Timeseries, Metric, Rollup and MetricScope to all be in the same
file, and to not be Expressions. All of these classes return complex types when
being serialized, in that they affect multiple parts of the query separately.
For example, Rollup impacts the conditions as well as the ORDER BY clause. So
each of these classes also has a visitor in a similar way to the query.

### DIFF
--- a/snuba_sdk/timeseries.py
+++ b/snuba_sdk/timeseries.py
@@ -13,18 +13,6 @@ class InvalidTimeseriesError(Exception):
     pass
 
 
-VALID_ENTITIES = set(
-    [
-        "generic_metrics_distributions",
-        "generic_metrics_counters",
-        "generic_metrics_sets",
-        "metrics_distributions",
-        "metrics_counters",
-        "metrics_sets",
-    ]
-)
-
-
 @dataclass(frozen=True)
 class Metric:
     """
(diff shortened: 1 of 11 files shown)


snuba_sdk/timeseries.py
```diff
@@ -1,15 +1,32 @@
...
-from snuba_sdk.expressions import Expression, InvalidExpressionError, is_literal
+from snuba_sdk.expressions import InvalidExpressionError, is_literal, list_type
+from snuba_sdk.orderby import Direction
+
+
+class InvalidTimeseriesError(Exception):
+    pass
+
+
+VALID_ENTITIES = set(
```
enochtangg, line 16: Some thoughts on entity validation. After this change, we would have snuba entity names defined in 3 places: sentry, snuba-sdk, and snuba itself. Entity resolution currently takes place in sentry (which although hardcoded, will always return a valid entity name). Do you think validation here is necessary? Since entity names are a snuba concept anyways and we aren't directly exposing these names outside (via some API), could we maybe even validate on the API side?

snuba_sdk/timeseries.py
```diff
@@ -18,17 +35,31 @@ class Metric(Expression):
     public_name: str | None = None
     mri: str | None = None
     id: int | None = None
+    entity: str | None = None
```
enochtangg, line 38: Could we add a `set_entity()` method for resolving the entity name?

enochtangg (commented):
Looks great, love the comprehensive tests. A few comments.

snuba_sdk/timeseries.py
evanh: Yeah good point. I'll remove it.

enochtangg (approved):

PR #126, opened 2023-10-19
feat(metrics): Introduce MQL grammar and create parsing function

enochtangg:
### Overview
This PR is responsible for adding a new `parse_mql()` function used to convert an MQL (Metrics Query Language) string into a MetricsQuery object. This is one of the first steps towards unifying the language we use at Sentry to query for metrics.  Using [Parsimonious](https://github.com/erikrose/parsimonious#parsimonious), a new grammar is created to express metric queries as strings. Then, with node visitors, we parse the string into a valid MetricsQuery object that is usable in the new Sentry metrics layer.

### Considerations
This parser currently only accepts simple timeseries and total queries. Though this parser understands formula queries (queries with arithmetic functions), the new metrics layer and snuba does not support that yet. As a result, this parser will fail those queries. See `tests/test_dsl.py` for examples of valid queries.


snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,296 @@
...
+        method = getattr(self, "visit_" + node.expr_name, self.generic_visit)
+        try:
+            result = method(node, [self.visit(n) for n in node])
+            return result
+        except Exception:
+            raise Exception
+
+    def collapse_into_timeseries(self, metric_query: MetricsQuery) -> Timeseries:
+        """
+        Collpases the filters and groupbys of a MetricsQuery into the Timeseries object.
```
evanh, line 115:
```suggestion
        Collapses the filters and groupbys of a MetricsQuery into the Timeseries object.
```

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,296 @@
...
+        ...the ``visit_bold()`` method would be called.
+        """
+        method = getattr(self, "visit_" + node.expr_name, self.generic_visit)
+        try:
+            result = method(node, [self.visit(n) for n in node])
+            return result
+        except Exception:
+            raise Exception
+
+    def collapse_into_timeseries(self, metric_query: MetricsQuery) -> Timeseries:
```
evanh, line 113: Why do you do this? Why not leave the filters/groupby at the `MetricsQuery` level? We'll end up with duplicated conditions. And when the query is sent to Snuba later, it will still have to push all the top level fields down to the timeseries again, in case any new timeseries were added/changed.

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,296 @@
...
+            bold = '<b>'
+
+        ...the ``visit_bold()`` method would be called.
+        """
+        method = getattr(self, "visit_" + node.expr_name, self.generic_visit)
+        try:
+            result = method(node, [self.visit(n) for n in node])
+            return result
+        except Exception:
+            raise Exception
```
evanh, line 111: This doesn't feel helpful. Why return a generic exception?

snuba_sdk/conditions.py
```diff
@@ -142,7 +142,7 @@ class Or(BooleanCondition):
     conditions: ConditionGroup = field(default_factory=list)
 
 
-ConditionGroup = Sequence[Union[BooleanCondition, Condition]]
+ConditionGroup = List[Union[BooleanCondition, Condition]]
```
evanh, line 145: If you're going to do this then please use `list` instead of `List`.

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,296 @@
...
+from snuba_sdk.column import Column, Variable
+from snuba_sdk.conditions import Condition, Op
+from snuba_sdk.function import Function
+from snuba_sdk.metrics_query import MetricsQuery
+from snuba_sdk.query_visitors import InvalidQueryError
+from snuba_sdk.timeseries import Metric, Timeseries
+
+
+AGGREGATE_PLACEHOLDER_NAME = "AGGREGATE_PLACEHOLDER"
+ENTITY_TYPE_REGEX = r"(c|s|d|g|e)"
```
evanh, line 22: I know that right now these values line up with Entities, but in the future they might not. Really these are metric types, not entity types.

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,296 @@
...
+NAMESPACE_REGEX = (
```
evanh, line 23: I don't know that we want to have this list here. Then every time someone wants to add a new use case they have to release a new version of the SDK.

I think it would be better if the SDK just accepted any string for the namespace. Then we could validate it in the Sentry code.

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,296 @@
...
+            target = target.set_groupby(group_by)
+        return target
+
+    def visit_condition(self, node: Node, children: Sequence[Any]) -> Any:
+        lhs, _, op, _, rhs = children
+        return Condition(lhs[0], op, rhs)
+
+    def visit_function(self, node: Node, children: Sequence[Any]) -> Any:
+        """
+        Given an target (which could be either a MetricsQuery or Timeseries object),
```
evanh, line 202: This is confusing me a bit: why can this be anything other than a Timeseries object? A MetricsQuery could be composed of multiple Timeseries, but I don't think we want to support a MetricsQuery being composed of multiple MetricsQuerys.

snuba_sdk/function.py
```diff
@@ -81,7 +88,8 @@ def validate(self) -> None:
                 )
             for param in self.parameters:
                 if not isinstance(
-                    param, (Column, CurriedFunction, Function, Identifier, Lambda)
+                    param,
+                    (Column, CurriedFunction, Function, Identifier, Lambda, Expression),
```
evanh, line 92: I'm guessing you added `Expression` to this to support Functions of Timeseries. However I don't want to add the Expression class here. All of these other classes are subclasses of Expression, the idea is to be restrictive of which specific ones are allowed in Functions.

snuba_sdk/timeseries.py
```diff
@@ -75,7 +76,7 @@ def set_entity(self, entity: str) -> Metric:
 
 
 @dataclass
-class Timeseries:
+class Timeseries(Expression):
```
evanh, line 79: I explictly didn't do this, because an `Expression` is assumed to be valid when it's instantiated (it runs `.validate()` in a `__post_init` function. Since Timeseries are complicated objects, I didn't want to enforce that they had to be valid immediately, so they could be constructed incrementally.

That's the reason this, `MetricsQuery` and `Query` are not instances of `Expression`.

test-requirements.txt
```diff
@@ -1,7 +1,9 @@
-pytest==6.1.2
+pytest==6.2.5
 pytest-forked==1.3.0
 pytest-localserver==0.5.0
 pytest-cov==2.10.1
+parsimonious==0.10.0
```
evanh, line 5: The last time we upgraded parsimonious past 0.8.1, we noticed a big performance hit. However I think that was version 0.9.x, so maybe 0.10.0 resolved it?

I think you should do a performance test between what we use in Snuba (0.8.1) and 0.10.0. If we can start upgrading parsimonious again I'd be pretty happy.

snuba_sdk/column.py
```diff
@@ -72,3 +72,8 @@ def validate_data_model(self, entity: Entity) -> None:
             raise InvalidColumnError(
                 f"entity '{entity.name}' does not support the column '{self.name}'"
             )
+
+
+@dataclass(frozen=True)
+class Variable(Column):
```
evanh, line 78: Why did you add this class?

tests/test_dsl.py
```diff
@@ -0,0 +1,377 @@
...
+from snuba_sdk.column import Column
+from snuba_sdk.conditions import Condition, Op
+from snuba_sdk.dsl.dsl import parse_mql
+from snuba_sdk.function import Function
+from snuba_sdk.metrics_query import MetricsQuery
+from snuba_sdk.timeseries import Timeseries, Metric
+
+
+def test_quoted_mri_name() -> None:
+    dsl = "sum(`d:transactions/duration@millisecond`)"
```
evanh, line 12: I'm not going to make you change all these, but I've been trying to use table tests for most of the SDK. I think these would be more concise in that format (https://github.com/getsentry/snuba-sdk/blob/main/tests/test_metrics_query.py#L16)

evanh (commented):
Let me know if you want to talk about problem solving around using the `Function` object inside the `MetricsQuery` object, since I think it might cause import loops?

snuba_sdk/conditions.py
enochtangg: Why is that? Using `list` produces mypy error: `list is not subscriptable`. That is because the expression `list[..]` is trying to subscript the object `list`

snuba_sdk/dsl/dsl.py
enochtangg: Oops copy pasta

snuba_sdk/dsl/dsl.py
enochtangg: Renamed to `METRIC_TYPE_REGEX`

snuba_sdk/dsl/dsl.py
enochtangg: Sounds good to me.

snuba_sdk/dsl/dsl.py
enochtangg: You're right, a MetricsQuery cannot be composed of multiple MetricQueries. After some more testing, it looks like target will actually only ever be a MetricsQuery. This is because if we step down the tree from the `function` node, one of its children is actually an `expression` node which resolves into a `MetricQuery`. If that's unclear, it might be easier to walk through it offline  and discuss alternatives.

snuba_sdk/function.py
enochtangg: Yes, that was my intention. I initially tried it with `Timeseries`, but that would make both modules require each other.

snuba_sdk/timeseries.py
enochtangg: Makes sense, thanks for pointing this out.

snuba_sdk/column.py
enochtangg: Was playing around with adding support for variables in expressions (e.g. `$foo`). Will remove for this PR.

snuba_sdk/conditions.py
evanh: You might have to add `from __future__ import annotations` to the top of the file. Using the lower case version is the default in newer versions of Python, so this is just making us forward compatible.

snuba_sdk/dsl/dsl.py
evanh: Ah I see, yes that does make sense.

I think this might be solved by not resolving things into a MetricsQuery until the entire DSL has been parsed. That way each component of the string is always guaranteed to return a Timeseries. Only after the entire string has been parsed into a Function or a Timeseries does it get wrapped up into a MetricsQuery.

enochtangg: As discussed offline, we may need to rework how the MetricsQuery, Timeseries, and Function classes look like. Currently, the MetricsQuery contains information about the top level filters and groupbys which requires it to resolve before the entire DSL is parsed given the way the nodes are traversed. We should move these top-level attributes elsewhere such that parsing the DSL resolves into a singular Timeseries/Equation.

enochtangg: Added in https://github.com/getsentry/snuba-sdk/pull/127.

snuba_sdk/function.py
enochtangg: Resolved with `Formula` class

snuba_sdk/timeseries.py
enochtangg: Resolved with Formula class

tests/test_dsl.py
enochtangg: Updated.

test-requirements.txt
enochtangg: Parsimonious 0.8.1 is unsupported for python 3.11.

Errors with:
```
ImportError: cannot import name 'getargspec' from 'inspect' (/opt/homebrew/Cellar/python@3.11/3.11.6/Frameworks/Python.framework/Versions/3.11/lib/python3.11/inspect.py)
```
because of deprecated `getargspec()`. Should use `inspect.signature()` or `inspect.getfullargspec()` instead.

snuba_sdk/conditions.py
enochtangg: Added, but it seems to be failing CI

test-requirements.txt
evanh: We use Python 3.8 in prod I believe. There's a push to get to 3.10 but it's not there yet.

This is good to know though. If the perf tests are OK we should upgrade.

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+                    combined_groupby = combined_groupby if combined_groupby else None
+                    new_timeseries = parameter.set_filters(combined_filters).set_groupby(combined_groupby)
+                    new_parameters.append(new_timeseries)
+                else:
+                    new_parameters.append(parameter)
+            formula.parameters = new_parameters
+
+
+
+    def visit_expression(self, node: Node, children: Sequence[Any]) -> Any:
```
evanh, line 148: All of these functions should be typed specifically. See the [SNQL parser](https://github.com/getsentry/snuba/blob/master/snuba/query/snql/parser.py#L325) for how we do that. We also type the children sequence as much as possible.

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+        return expr
+
+    def visit_expr_op(self, node: Node, children: Sequence[Any]) -> Any:
+        raise InvalidQueryError("Arithmetic function not supported yet")
+        return EXPRESSION_OPERATORS[node.text]
+
+    def visit_term(self, node: Node, children: Sequence[Any]) -> Any:
+        """
+        Checks if the current node contains two term children, if so
+        then merge them into a single Function with the operator. If the
```
evanh, line 162:
```suggestion
        then merge them into a single Formula with the operator. If the
```

snuba_sdk/dsl/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+    def visit_variable(self, node: Node, children: Sequence[Any]) -> Any:
+        raise InvalidQueryError("Variables are not supported yet")
+        return None
+
+    def visit_nested_expression(self, node: Node, children: Sequence[Any]) -> Any:
+        return children[2]
+
+    def visit_aggregate(self, node: Node, children: Sequence[Any]) -> Any:
+        """
+        Given a target (which is either a MetricsQuery or Timeseries object),
```
evanh, line 277:
```suggestion
        Given a target (which is either a Formula or Timeseries object),
```

snuba_sdk/expressions.py
```diff
@@ -26,7 +26,7 @@ def validate(self) -> None:
 # For type hinting
 ScalarLiteralType = Union[None, bool, str, bytes, float, int, date, datetime]
 ScalarSequenceType = Sequence[Union[Expression, ScalarLiteralType]]
-ScalarType = Union[ScalarLiteralType, ScalarSequenceType]
+ScalarType = Union[ScalarLiteralType, ScalarSequenceType, Expression]
```
evanh, line 29: We shouldn't need this any more, and if we do there's a deeper issue we should unravel.

snuba_sdk/function.py
evanh: Shouldn't this be removed?

snuba_sdk/timeseries.py
evanh: Shouldn't this be removed now?

evanh (commented):
Looking really good!

A few nits here and there.

Also, can you please update CHANGES.rst?

snuba_sdk/conditions.py
enochtangg: Using `list` only works for python > 3.9. We are still on 3.8 in prod so unfortunately we still need to use `List` from typing

test-requirements.txt
evanh: Perfect, thanks for that! Would be good to link to any results you have.

evanh (approved):
Pending the CI passing, this looks good!

PR #127, opened 2023-10-25
feat(metrics): Add formula class

enochtangg:
This PR is responsible for adding a new `Formula` class (inherits Expression) which will be used to replace `Function` in MetricsQuery objects. Upon implementing the first iteration of the DSL parser (https://github.com/getsentry/snuba-sdk/pull/126), it was discovered that MetricsQuery, Timeseries and Function classes are not defined intuitively with relation to the new DSL grammar and how the nodes are parsed. Practically, it helps to reason about queries by having arbitrary levels of groupbys and filters. Although, in the AST itself, all filters and groupbys will be pushed down. This class aims to simplify the complexity of the `MetricsQuery` by moving those attributes here.

### DIFF
--- a/snuba_sdk/formula.py
+++ b/snuba_sdk/formula.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+from enum import Enum
+from dataclasses import dataclass
+from typing import Optional, Sequence, Union
+
+from snuba_sdk.column import Column
+from snuba_sdk.expressions import Expression, InvalidExpressionError
+from snuba_sdk.aliased_expression import AliasedExpression
+from snuba_sdk.conditions import ConditionGroup
+from snuba_sdk.timeseries import Timeseries
+
+
+class InvalidArithmeticError(InvalidExpressionError):
+    pass
+
+
+class InvalidFormulaError(InvalidExpressionError):
+    pass
+
+
+class ArithmeticOperator(Enum):
+    PLUS = "plus"
+    MINUS = "minus"
+    MULTIPLY = "multiply"
+    DIVIDE = "divide"
+
+
+@dataclass(frozen=True)
+class Formula(Expression):
+    operator: str
+    parameters: Optional[
+        Sequence[
+            Union[Formula, Timeseries, float, int]
+        ]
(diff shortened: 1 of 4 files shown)


snuba_sdk/formula.py
```diff
@@ -0,0 +1,88 @@
...
+class ArithmeticOperator(Enum):
+    PLUS = "plus"
+    MINUS = "minus"
+    MULTIPLY = "multiply"
+    DIVIDE = "divide"
+
+
+@dataclass(frozen=True)
+class Formula(Expression):
+    operator: str
```
evanh, line 31: This should be an instance of `ArithmeticOperator` instead of `str`. Then we can be sure it's an allowed operation.

snuba_sdk/formula.py
```diff
@@ -0,0 +1,88 @@
...
+                raise InvalidFormulaError(
+                    f"parameters of formula {self.operator} must be a Sequence"
+                )
+            for param in self.parameters:
+                if not isinstance(param, (Formula, Timeseries, float, int)):
+                    raise InvalidFormulaError(
+                        f"parameter '{param}' of formula {self.operator} is an invalid type"
+                    )
+
+    def __eq__(self, other: object) -> bool:
```
evanh, line 61: I don't think you need this. `dataclass` automatically does this.

snuba_sdk/formula.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ArithmeticOperator(Enum):
+    PLUS = "plus"
+    MINUS = "minus"
+    MULTIPLY = "multiply"
+    DIVIDE = "divide"
+
+
+@dataclass(frozen=True)
+class Formula(Expression):
```
evanh, line 30: If this can be modified (with the `set_*`) functions, then it shouldn't extend the Expression class. You'll have to manually call validate on it in the MetricsQuery validator visitor.

enochtangg: Removed `Expression` and updated validator visitor to take Formula type.

evanh (approved):

PR #128, opened 2023-10-25
ref(docs): Switch from Sphinx to pdoc3

evanh:
Sphinx has a persistent bug in every version tested (including on Python 3.9):
https://github.com/sphinx-doc/sphinx/issues/9243

There seems to be no documented fix for this. Instead of fighting with this,
switch to a simpler documentation engine, [pdoc3](https://pdoc3.github.io/pdoc/doc/pdoc/#gsc.tab=0). This has definitely less
features than Sphinx but is way easier to use.

Screenshot 2023-10-25 at 5 46 13 PM

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,17 +19,17 @@ jobs:
       - name: Build docs
         run: |
           pip install virtualenv
-          make apidocs
+          make generate-docs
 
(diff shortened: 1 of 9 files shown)


evanh:
This draft doesn't actually switch it over yet, I can add that if people are OK with the change.

volokluev (approved):

PR #131, opened 2023-10-30
ref(metrics): Support offset and limit in MetricsQuery

evanh:
In order to properly paginate queries, it's necessary to be able to specify and
offset and a limit to queries. Add these fields to the MetricsQuery, which then
get added to the appropriate LIMIT/OFFSET clauses in SnQL.


enochtangg (approved):
Will need to change to 2.0.8 since I merged MQL feature already.

PR #132, opened 2023-11-01
fix(metrics): Fix Timeseries Filters Validator

enochtangg:
The Timeseries `filter` property should be able to have type `Condition` and/or `BooleanCondition`. This PR is responsible for fixing that.


volokluev (commented):
Test?

enochtangg:
yes thank you

evanh (approved):

PR #133, opened 2023-11-03
feat(metrics): Add Experimental Formula Query Support

enochtangg:
### Overview
This PR is explores what needs to be done to support metric formula queries without needing to change the Snuba API. This is not meant to be a permanent solution as [creating a new Snuba endpoint which accepts a DSL string and parses the formula directly into an AST is the end goal](https://www.notion.so/sentry/Metrics-Query-Layer-91fc5a1e11274019a0d51985f7f91892?pvs=4#a3e4b8dbf3c844dc8c17cbf4d6b1b4cd). This PR is meant to explore how formula queries would work without needing Snuba changes and possibly grab some low hanging fruit that might make formula queries operational immediately.

This PR explores the following changes:
1. Decouples the SnQL string from the `SnQLPrinter` which is associated to one MetricsQuery. This is because with the introduction of Formula queries, it is possible to have one or more SnQL queries associate to one MetricsQuery. For example `sum(d:foo) / sum(d:bar) by transactions` requires two SnQL queries. To decouple this, we introduce `SnQLString` and `FormulaSnQL` class which get return from `SnQLPrinter._combine()`.
2. To support formula queries, a `FormulaSnQLPrinter` class was added which is responsible for recursively visiting its parameters and ultimately converting a `Formula` into a a mapping which can be transformed into a `FormulaSnQL`.
3. The `FormulaSnQLPrinter` is also responsible for collapsing all context attributes into the inner timeseries. For example:
```
`(sum(d:foo) / sum(d:bar)){tag: value} by transactions * 1000`
becomes
`(sum(d:foo{tag: value}) by transactions  / sum(d:bar{tag: value}) by transactions ) * 1000`
```
Similarly, attributes like start, end, rollup, limit, and etc. all get pushed into the inner timeseries SnQL. See `test_metrics_query.py::test_formula_query`

### What changes will be required in sentry?
Non-metric queries and timeseries metric queries will still serialize as normal. The serialization of a Formula within a MetricsQuery within a Request object is the complicated part. This is because serializing a Formula would generate multiple SnQLs which should produce multiple Requests. It feels like a strange pattern to call `serialize` on one request which generates multiple serialize requests. This still needs to be explored. For now, in the `MetricsQuery` class, we introduce a new method called `serialize_formula()` which returns a `FormulaSnQL`. On the sentry side, we would need to alter both the new metrics layer and `sentry/utils/snuba.py::_raw_snql_query()` to know when to call the new method, run the multiple snuba queries, and compute the arithmetics.

### DIFF
--- a/snuba_sdk/dsl/dsl.py
+++ b/snuba_sdk/dsl/dsl.py
@@ -124,7 +124,6 @@ class MQLlVisitor(NodeVisitor):  # type: ignore
         return expr
 
     def visit_expr_op(self, node: Node, children: Sequence[Any]) -> Any:
-        # raise InvalidQueryError("Arithmetic function not supported yet")
         return EXPRESSION_OPERATORS[node.text]
 
     def visit_term(
@@ -142,7 +141,6 @@ class MQLlVisitor(NodeVisitor):  # type: ignore
         return term
 
     def visit_term_op(self, node: Node, children: Sequence[Any]) -> Any:
-        # raise InvalidQueryError("Arithmetic function not supported yet")
         return TERM_OPERATORS[node.text]
 
     def visit_coefficient(
--- a/snuba_sdk/formula.py
+++ b/snuba_sdk/formula.py
@@ -83,11 +83,10 @@ class Formula:
 FormulaParameterGroup = Union[Formula, Timeseries, float, int]
 
 
-@dataclass(frozen=True)
 class FormulaSnQL:
     """
     Temporary class to represent a Formula in SnQL. This will be removed once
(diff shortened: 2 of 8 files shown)


evanh (commented):
The fact that this creates multiple snuba queries is a problem. If there isn't a way to to build a single SnQL query, then I think this strategy is a no-go.

enochtangg:
@evanh Agreed. The serialization (serializing one `Request` into multiple `Request`) is the biggest problem and would require pretty big changes in sentry. This isn't really worth to further explore given our current constraints.

PR #134, opened 2023-11-06
feat(mql): Translate MetricsQuery to MQL string

enochtangg:
This PR is responsible for adding `MQLPrinter`, `TimeseriesMQLPrinter`, and `TranslationMQL` classes which are necessary for converting a `MetricsQuery` (only timeseries, no formula queries yet) into its MQL string form. To convert a MetricsQuery into MQL string, a `serialize_to_mql()` method was added to the class. 

To handle the serialization of the MQL context attributes (start, end, rollup, scope, limit, offset, etc.), an optional field called `mql_context` was introduced to the `Request` object. There are two places which populate this field. First, in the sentry metrics layer, a `MetricsQuery` object is first resolved with the indexer and a mapping is created. This mapping gets set on the Request object. Later on, when the `Request` gets serialized, the snuba-sdk populates the rest of the fields in the `mql_context` (everything else that isn't `indexer_mappings`). 

The final serialized MetricQuery in MQL form would look something like:
```
{
	"query": "max(d:transactions/duration@millisecond){bar:'baz'}",
	"mql_context": {
		"entity": "generic_metrics_distributions"
		"start": "2023-01-02T03:04:05+00:00",
		"end": "2023-01-16T03:04:05+00:00",
		"rollup": {
			"orderby": {"column_name": "time", "direction": "ASC"},
			"granularity": "3600",
			"interval": "3600",
			"with_totals": "",
		},
		"scope": {
			"org_ids": [1],
			"project_ids": [11],
			"use_case_id": "transactions",
		},
		"limit": "",
		"offset": "",
		"indexer_mappings": {"d:transactions/duration@millisecond": "123456", ...}
	}
	"dataset": ...
	"app_id": ...
	"tenant_ids": ...
	"parent_api": ...
}
```


enochtangg:
Updated this PR to align with discover filter grammar (https://github.com/getsentry/snuba-sdk/pull/135)

snuba_sdk/metrics_visitors.py
```diff
@@ -148,6 +208,18 @@ def visit(self, metric: Metric) -> Mapping[str, str]:
         }
 
 
+class MetricMQLPrinter(MetricVisitor[Mapping[str, str]]):
+    def visit(self, metric: Metric) -> Mapping[str, str]:
```
evanh, line 212: This requires more thought. We need the id in the MQL query, at least until we have the indexer working in Snuba. Otherwise Snuba won't be able to actually run the query.

snuba_sdk/visitors.py
```diff
@@ -264,6 +264,95 @@ def _visit_totals(self, totals: bool) -> str:
...
+    def _visit_join(self, join: Join) -> str:
+        raise NotImplementedError
+
+    def _visit_condition(self, cond: Condition) -> str:
+        rhs = None
+        if is_unary(cond.op):
+            rhs = ""
+        elif isinstance(cond.rhs, (CurriedFunction, Function)):
+            raise InvalidExpressionError(
+                "CurriedFunction and Function are supported in conditions in MQL"
```
evanh, line 330:
```suggestion
                "CurriedFunction and Function are not supported in conditions in MQL"
```

snuba_sdk/visitors.py
```diff
@@ -264,6 +264,95 @@ def _visit_totals(self, totals: bool) -> str:
         return f"{totals}"
 
 
+class TranslationMQL(ExpressionVisitor[str]):
```
evanh, line 267: If the only thing that's different about this is just a couple functions, then I would extend the existing `Translation` class and override just the changed functions.

snuba_sdk/visitors.py
```diff
@@ -264,6 +264,95 @@ def _visit_totals(self, totals: bool) -> str:
...
+    def __init__(self, use_entity_aliases: bool = False):
+        self.use_entity_aliases = use_entity_aliases
+
+    def _stringify_scalar(self, value: ScalarType) -> str:
```
evanh, line 271: Why did you change this?

enochtangg: The `_stringify_scalar` in the `Translation` class does a few extra things that are SnQL specific. For example, it prepends `array` and `tuple` to certain scalars which we don't want for MQL.

snuba_sdk/metrics_visitors.py
enochtangg: I agree, but I don't think we should encode the ID into the MQL query as a temporary solution, since we can already query snuba with the new metrics layer as it exists right now. I think moving the indexer to Snuba should be a requirement in order for the new snuba endpoint to function.

snuba_sdk/visitors.py
evanh: Fair enough.

snuba_sdk/metrics_visitors.py
evanh: I don't want to block work on this layer behind the indexer work. I would rather build this in a way that supports IDs, then remove that support once the indexer work is done.

We don't know how long that will take or how easy it will be, and blocking metrics layer development with it I think will end up hampering us in the future.

What we should do is call this out as tech debt that we need to clean up once the indexer work is done.

enochtangg: Sure that makes sense. Let's chat about how we might want to do this offline, because I think there's an opportunity to group this data together with the other context fields (scope, limit, totals, start, end, etc.) cleanly.

evanh (commented):
I like this, I have one suggestion though:

I don't think the `mql_context` should encode query strings the way it does. For example for `start` and `end`, it should just encode the datetime itself. That can be done similarly for all the other fields as well. For the scope, encode the project_ids and org_ids and use_cases in a dictionary `{"project_ids": [...], "org_id": ...}`.

The reason for this is that in Snuba, we deal with AST objects. In order to convert the `mql_context` into the AST object, we will now have to parse the SnQL encoded strings to get the raw information. It would be easier to have just the start and end timestamps, and add a condition object to the AST, then to decode the condition string, and then add that to the AST.

evanh (approved):

PR #135, opened 2023-11-07
ref(MQL): Update MQL Grammar to align with Discover Filter Grammar

enochtangg:
### Overview
The goal is to eventually unify all our metrics with a single method of querying. Right now, how we express filters is different between discover and metrics. We want to close this gap in order to provide a smooth experience across our product. This PR is responsible for updating the MQL grammar to more closely match how discover exposes filters to our users.

### What is not included in this PR
* OR filtering
* bracket filters


evanh (approved):
Another thing we are losing with this change is the wildcard search support (`~` and `LIKE`). Discover does that using wildcard strings `status_code:"5*"`.

PR #136, opened 2023-11-08
ref(formula) Add SnQL visitor for Formula

evanh:
First off, note that this is a temporary measure. Formulas don't encode nicely
into SnQL. The main issue with encoding MetricsQuery into SnQL is that the
visitor pattern assumes each component of a query can be encoded into a string,
and then combined together neatly into a query. That's not true of a
MetricsQuery, and especially a Formula, where the encoding of a single
component can impact multiple parts of a query. For example encoding a Metric
object influences the WHERE clause (metric_id filter) and the MATCH clause (for
the entity).

This assumption works much better when encoding MetricsQuery into MQL, since
the components can be encoded and combined together in a seamless way. Until
then, these visitors will have to hack around the restrictions.

This adds a simple SnQL visitor for the Formula class. This only works for the
most basic Formulas: no nested Formulas as allowed, and the groupby and entity
of any Timeseries parameters must match.

### DIFF
--- a/snuba_sdk/formula.py
+++ b/snuba_sdk/formula.py
@@ -62,9 +62,10 @@ class Formula:
                             "Formulas can only operate on a single entity"
                         )
 
+                    to_check = set(param.groupby) if param.groupby is not None else None
                     if groupby is None:
-                        groupby = param.groupby
-                    elif groupby != param.groupby:
+                        groupby = to_check
+                    elif groupby != to_check:
                         raise InvalidFormulaError(
                             "Formula parameters must group by the same columns"
                         )
--- a/snuba_sdk/formula.py
+++ b/snuba_sdk/formula.py
@@ -43,11 +43,34 @@ class Formula:
                 raise InvalidFormulaError(
                     f"parameters of formula {self.operator.value} must be a Sequence"
                 )
+
+            # - Must have the same entities, and must have an entity (no formulas on just literals)
+            # - Must have the same groupbys
+            entity = None
+            groupby = None
             for param in self.parameters:
-                if not isinstance(param, (Formula, Timeseries, float, int)):
+                if not isinstance(param, (Timeseries, float, int)):
                     raise InvalidFormulaError(
                         f"parameter '{param}' of formula {self.operator.value} is an invalid type"
                     )
+
+                if isinstance(param, Timeseries):
+                    if entity is None:
+                        entity = param.metric.entity
+                    elif entity != param.metric.entity:
+                        raise InvalidFormulaError(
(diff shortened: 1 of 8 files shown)


snuba_sdk/formula.py
```diff
@@ -43,12 +43,35 @@ def validate(self) -> None:
...
+                    if entity is None:
+                        entity = param.metric.entity
+                    elif entity != param.metric.entity:
+                        raise InvalidFormulaError(
+                            "Formulas can only operate on a single entity"
+                        )
+
+                    if groupby is None:
+                        groupby = param.groupby
+                    elif groupby != param.groupby:
```
enochtangg, line 67: Can we check for the equality of sets here in case they contain the same elements but out of order?

tests/test_formula.py
```diff
@@ -98,3 +215,161 @@ def test_formulas(
         with pytest.raises(type(exception), match=re.escape(str(exception))):
             formula = func_wrapper()
             formula.validate()
+
+
+formula_snql_tests = [
```
enochtangg, line 220: Up to you, but I think it'd help me better understand which formula queries aren't supported if we add invalid tests (e.g. different groupbys among timeseries, etc.)

snuba_sdk/metrics_visitors.py
```diff
@@ -229,3 +236,141 @@ def visit(self, scope: MetricsScope) -> str:
...
+        filters = []
+        if timeseries.filters:
+            for cond in timeseries.filters:
+                assert isinstance(cond.op, Op)  # BooleanOp is not a valid function
+                op = OPERATOR_TO_FUNCTION[cond.op]
+                filters.append(Function(op.value, [cond.lhs, cond.rhs]))
+
+        and_filters = Function("and", [metric_filter, *filters])
+        aggregate = CurriedFunction(
+            f"{timeseries.aggregate}If",
```
enochtangg, line 351: if we're doing this, should we validate the aggregate to some allowlist? This also opens up the question: which aggregation function do we want to support out the box (if not all) and should they be the same across all entities?

enochtangg (commented):
Excited for this! Just have a few questions.

snuba_sdk/metrics_visitors.py
evanh: I do think we need an allow list, but I would do that in a separate PR from this.

tests/test_formula.py
evanh: I added those to the formula tests above. These tests are now simply testing the serialization. The regular formula tests are testing all the cases I added.

enochtangg (approved):

PR #141, opened 2023-11-21
feat(mql): Support unquoted tag values and space delimiters in MQL filter

enochtangg:
This PR is responsible for extending the MQL grammar to support unquoted tag values and the usage of spaces as delimiters in the filter string. The purpose of doing this is to bring MQL closer to the discover grammar in order to provide a cohesive user experience across the sentry product.

### DIFF
--- a/snuba_sdk/dsl/dsl.py
+++ b/snuba_sdk/dsl/dsl.py
@@ -38,8 +38,8 @@ condition_op = "!"
 tag_key = ~r"[a-zA-Z0-9_]+"
 tag_value = quoted_string / unquoted_string / string_tuple / variable
 
-quoted_string = quote unquoted_string quote
-unquoted_string = ~r"[a-zA-Z0-9_]+"
+quoted_string = ~r'"([^"\\]*(?:\\.[^"\\]*)*)"'
+unquoted_string = ~r'[^,\[\]\"}}{{\s]+'
 string_tuple = open_square_bracket _ (quoted_string / unquoted_string) (_ comma _ (quoted_string / unquoted_string))* _ close_square_bracket
 
 target = variable / nested_expression / function / metric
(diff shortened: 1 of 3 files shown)


snuba_sdk/dsl/dsl.py
```diff
@@ -31,33 +31,34 @@
...
 condition = condition_op? (variable / tag_key) _ colon _ tag_value
 condition_op = "!"
-tag_key = ~"[a-zA-Z0-9_]+"
-tag_value = quoted_string / quoted_string_tuple / variable
+tag_key = ~r"[a-zA-Z0-9_]+"
+tag_value = quoted_string / unquoted_string / string_tuple / variable
 
-quoted_string = ~r'"([^"\\]*(?:\\.[^"\\]*)*)"'
-quoted_string_tuple = open_square_bracket _ quoted_string (_ comma _ quoted_string)* _ close_square_bracket
+quoted_string = quote unquoted_string quote
```
evanh, line 41: This needs to be changed back to the more open regex above.

The quoted string regex needs to match all special characters and weird stuff, that's the purpose of the quotes.

snuba_sdk/dsl/dsl.py
```diff
@@ -31,33 +31,34 @@
...
+unquoted_string = ~r"[a-zA-Z0-9_]+"
```
evanh, line 42: Related to the comment above, this is too restrictive. This regex will fail if someone tries to match `release_data=2023-01-03T10:00:00` because you aren't allowing dashes or colons. I would recommend, instead of trying to whitelist all the possible allowed characters, maybe blacklist the ones the grammar cares about (commas, colons etc.)

enochtangg: Updated with blacklist characters.

evanh (approved):

PR #143, opened 2023-11-28
feat(mql) Support curried functions in MQL

evanh:
One of the main uses for metrics is currently quantiles. Those are curried
functions, so MQL needs to support curried functions if it is going to be used
to encode queries being sent to Snuba.

Add curried functions to the MQL grammar that support arbitrary params. The
only use case at the moment is quantiles but it doesn't hurt to be a little
more open in the grammar.

Also rename `dsl` to `mql` to be more consistent.
Also add `(` and `)` to the excluded characters in unquoted strings.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 Changelog and versioning
 ==========================
 
+2.0.12
+-----
+
+- Support curried functions in the MQL grammar
+
+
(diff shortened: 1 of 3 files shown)


enochtangg (approved):
Nice. I think this can be released with `2.0.11` since that version hasn't gone out yet.

PR #144, opened 2023-11-29
feat(metrics-layer): Add support for more complex filters grammar

iambriccardo:


enochtangg:
Thanks for adding this! Two things:

* If we are now supporting the `OR` filter, we should add tests to https://github.com/getsentry/snuba-sdk/blob/dff58ea1e4f5f7450658e09319cc0e7ab373a341/tests/test_metrics_query.py#L18 The translation from `MetricsQuery` to SnQL (for `OR` filter) should already work out the box.
* There should be follow up task to update the `TimeseriesMQLPrinter` which translates `MetricQuery` to MQL https://github.com/getsentry/snuba-sdk/blob/dff58ea1e4f5f7450658e09319cc0e7ab373a341/snuba_sdk/metrics_visitors.py#L131

enochtangg (commented):
Thanks for adding this! Two things:

* If we are now supporting the `OR` filter, we should add tests to https://github.com/getsentry/snuba-sdk/blob/dff58ea1e4f5f7450658e09319cc0e7ab373a341/tests/test_metrics_query.py#L18 The translation from `MetricsQuery` to SnQL (for `OR` filter) should already work out the box.
* There should be follow up task to update the `TimeseriesMQLPrinter` which translates `MetricQuery` to MQL https://github.com/getsentry/snuba-sdk/blob/dff58ea1e4f5f7450658e09319cc0e7ab373a341/snuba_sdk/metrics_visitors.py#L131

iambriccardo:
@enochtangg great feedback! Thanks

I can def add the support for the MQL generator and the tests in this PR, or you would prefer a separate PR?

enochtangg:
@iambriccardo same PR is fine, thanks again!

enochtangg (approved):

PR #148, opened 2023-12-05
fix(mql): Add support for `.` in the grammar

iambriccardo:
This PR adds the support for `.` in the MQL grammar to handle cases such as `sum(something) by (http.status_code)`.


enochtangg (approved):
Nice catch. Could we add small test case for this?

iambriccardo:
Yes, definitely!

PR #149, opened 2023-12-05
ref(MQL): Simplify MQLContext validation/serialization

evanh:
This arose from the realization that MQLContext is an internal class only. It
is only ever created when serializing a MetricsQuery, and that MetricsQuery
must be valid in order to be serialized in the first place.

Thus there's no need to do extra validation work of the MQLContext. Once that
was removed, the serializer also didn't seem necessary. The objects themselves
could be serialized and simply joined together in the MQLContext class.

The class itself is still useful, to provide a human readable understanding of
what fields are allowed in a serialized MQL query.

This also fixes a bug with the indexer_mappings field. That field will need to
be passed from Sentry to Snuba, in order for Snuba to do resolving. That was
added to the MetricsQuery class. In the future the resolving itself will happen
in Snuba, but for now this is required.


evanh:
I accidentally messed up the commit messages for this PR. I'll fix them when I merge.

snuba_sdk/mql_visitor.py
```diff
@@ -103,32 +112,41 @@ def _visit_query(self, query: Timeseries | Formula | None) -> Mapping[str, str]:
...
         if scope is None:
             raise InvalidMetricsQueryError("MetricQuery.scope must not be None")
 
-        return scope
+        return self.scope_visitor.visit(scope)
 
-    def _visit_limit(self, limit: Limit | None) -> Limit | None:
-        return limit
+    def _visit_limit(self, limit: Limit | None) -> int | None:
+        return limit.limit if limit is not None else None
```
enochtangg, line 140: If we're going to change this to allow `None`, we need to update the parser to reflect this: https://github.com/getsentry/snuba/pull/5119/files#diff-602995071cc68736a6c6f5eecb5672ec85419d9a80b9e3aba7ac963b098c70b6R694-R751

enochtangg (approved):
Nice, just had one note.

PR #151, opened 2023-12-11
fix(mql): Have MQL use double instead of single quotes

evanh:
MQL is meant to align with Discover syntax as much as possible. Discover uses
double quotes instead of single quotes, so when a MetricsQuery is serialized it
needs to be serialized using double quotes instead of single quotes.

### DIFF
--- a/snuba_sdk/visitors.py
+++ b/snuba_sdk/visitors.py
@@ -122,8 +122,9 @@ class ExpressionVisitor(ABC, Generic[TVisited]):
 
 
 class Translation(ExpressionVisitor[str]):
-    def __init__(self, use_entity_aliases: bool = False):
+    def __init__(self, use_entity_aliases: bool = False, is_mql: bool = False):
         self.use_entity_aliases = use_entity_aliases
+        self.is_mql = is_mql
 
     def _stringify_scalar(self, value: ScalarType) -> str:
         if value is None:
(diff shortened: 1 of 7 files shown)


volokluev:
Do we need to change the snuba code as well?

evanh:
> Do we need to change the snuba code as well?

@volokluev Yes, and I will open that PR once this is merged. I can't open an official PR until this is released.

snuba_sdk/visitors.py
```diff
@@ -122,8 +122,9 @@ def _visit_totals(self, totals: bool) -> TVisited:
 
 
 class Translation(ExpressionVisitor[str]):
-    def __init__(self, use_entity_aliases: bool = False):
+    def __init__(self, use_entity_aliases: bool = False, is_mql: bool = False):
```
volokluev, line 125: What can it be besides `mql`? In which cases would `is_mql` be False?

volokluev (commented):
Is it correct to say that the grammar already uses double quotes and we just need to change the mql generation?

snuba_sdk/visitors.py
evanh: This is used to encode both SnQL and MQL. In SnQL we use single quotes, `array(...)` and `tuple(...)`. In MQL we use double quotes and `[...]`.

volokluev (approved):

PR #152, opened 2023-12-12
fix(mql): The MQL serializer was not serializing curried aggregates

evanh:
Add functionality that properly serializes curried aggregates into MQL.


nikhars (approved):
LGTM

PR #155, opened 2024-01-02
feat(mql): Support arbitrary functions in MQL Grammar

enochtangg:
### Overview
This PR is responsible for supporting arbitrary functions names and parameters in MQL grammar. Since these functions are what the UI will interact with, they need to be encoded and parsed out of the MQL.
For example: `apdex(sum(transaction.duration), 500)` becomes

```
Formula(
    function_name="apdex",
    parameters=[
        Timeseries(metric=Metric("transaction.duration")),
        500
    ]
)
```

To accomplish this, `Formula.operator: ArithmeticOperator` was changed to `Formula.function_name: str` in order to extend the `Formula` class to handle arbitrary functions. This change also supports serializing back into MQL

### Testing
Unit tests for parsing MQL to MetricsQuery and MetricsQuery to MQL were added for arbitrary functions.

### DIFF
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -60,8 +60,8 @@ arbitrary_function = arbitrary_function_name (open_paren ( _ expression _ ) (_ c
 aggregate_list = param* (param_expression)
 param = param_expression _ comma _
 param_expression = number / quoted_string / unquoted_string
-aggregate_name = "avg" / "count" / "max" / "min" / "sum"
-curried_aggregate_name = "quantiles"
+aggregate_name = "avg" / "count" / "max" / "min" / "sum" / "last" / "uniq"
+curried_aggregate_name = "quantiles" / "histogram"
 arbitrary_function_name = ~r"[a-zA-Z0-9_]+"
 
 group_by = _ "by" _ (group_by_name / group_by_name_tuple)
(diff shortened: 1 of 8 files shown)


snuba_sdk/mql/mql.py
```diff
@@ -87,16 +90,26 @@
...
 
-TERM_OPERATORS: Mapping[str, ArithmeticOperator] = {
-    "*": ArithmeticOperator.MULTIPLY,
-    "/": ArithmeticOperator.DIVIDE,
+TERM_OPERATORS: Mapping[str, str] = {
+    "*": ArithmeticOperator.MULTIPLY.value,
+    "/": ArithmeticOperator.DIVIDE.value,
 }
 
+AGGREGATION_FUNCTIONS: set[str] = {
```
evanh, line 103: This is missing some aggregations. Offhand I can think of `quantiles` and `uniq` but there are others as well. I would go through all the entity definitions for metrics and generic_metrics to get everything. See for example here: https://github.com/getsentry/snuba/blob/master/snuba/datasets/configuration/generic_metrics/entities/distributions.yaml#L70.

tests/test_formula_printer.py
```diff
@@ -336,6 +322,116 @@
             groupby=[Column("transaction")],
         ),
         '((sum(foo){tag2:"tag_value2" AND (tag:"tag_value" OR tag:"tag_valueor")} / sum(bar)){tag3:"tag_value3"} * sum(pop)) by (transaction)',
+        id="test complex nested terms with OR condition",
+    ),
+    pytest.param(
```
evanh, line 327: Can you add a test somewhere (doesn't have to be this file) for a function on a quantile? e.g. `apdex(quantiles(0.5)(...`. Just to ensure that works?

snuba_sdk/mql/mql.py
enochtangg: Added missing aggregations. What do you think about them living directly on the grammar? It makes separating aggregates, curried aggregates, and arbitrary functions easier.

evanh (approved):

PR #158, opened 2024-01-12
ref(mql): Use `Formula` for aliased aggregates and arbitrary functions

enochtangg:
### Overview
This PR is responsible the following changes:
* Make aggregates require a metric as a parameter
* Allow string params by using `""`
* Uses `Formula` to capture any function that **is not** an aggregate function (arbitrary functions).
* Adding curried arbitrary function support (e.g. `topK(10)(transaction.duration)`).

### Changes in sentry
Need to update the alias resolver in sentry metrics layer code to support this.

### Testing
See added test cases for what is now supported.

### DIFF
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -383,7 +383,7 @@ class MQLVisitor(NodeVisitor):  # type: ignore
 
     def visit_curried_arbitrary_function(
         self, node: Node, children: Sequence[Any]
-    ) -> Formula:
+    ) -> Union[Timeseries, Formula]:
         """
         Returns a Fomula with the arbitrary curried function name and parameters.
         topK(10)(sum(mri)) -> Formula(topK, (Timeseries(sum), 10))
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -55,11 +55,10 @@ nested_expression = open_paren _ expression _ close_paren
 
 function = (curried_aggregate / curried_arbitrary_function / aggregate / arbitrary_function) (group_by)?
 aggregate = aggregate_name (open_paren _ inner_filter _ close_paren)
-inner_filter = metric (open_brace _ filter_expr _ close_brace)? (group_by)?
 aggregate_name = ~r"[a-zA-Z0-9_]+"
 arbitrary_function = arbitrary_function_name (open_paren ( _ expression _ ) (_ comma _ expression)* close_paren)
 arbitrary_function_name = ~r"[a-zA-Z0-9_]+"
-curried_aggregate = curried_aggregate_name (open_paren _ aggregate_list? _ close_paren) (open_paren _ filter _ close_paren)
+curried_aggregate = curried_aggregate_name (open_paren _ aggregate_list? _ close_paren) (open_paren _ inner_filter _ close_paren)
 curried_aggregate_name = "quantiles" / "histogram"
 curried_arbitrary_function = curried_arbitrary_function_name (open_paren _ aggregate_list? _ close_paren) (open_paren _ ( _ expression _ ) _ close_paren)
 curried_arbitrary_function_name = ~r"[a-zA-Z0-9_]+"
@@ -72,6 +71,7 @@ group_by = _ "by" _ (group_by_name / group_by_name_tuple)
 group_by_name = ~r"[a-zA-Z0-9_.]+"
 group_by_name_tuple = open_paren _ group_by_name (_ comma _ group_by_name)* _ close_paren
 
+inner_filter = metric (open_brace _ filter_expr _ close_brace)? (group_by)?
 metric = quoted_mri / unquoted_mri / quoted_public_name / unquoted_public_name
 quoted_mri = backtick unquoted_mri backtick
 unquoted_mri = ~r"{METRIC_TYPE_REGEX}:{NAMESPACE_REGEX}/{MRI_NAME_REGEX}@{UNIT_REGEX}"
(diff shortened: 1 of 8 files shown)


snuba_sdk/mql/mql.py
```diff
@@ -53,16 +52,17 @@
...
 
-function = (aggregate / curried_aggregate / arbitrary_function) (group_by)?
+function = (curried_aggregate / aggregate / arbitrary_function) (group_by)?
 aggregate = aggregate_name (open_paren _ filter _ close_paren)
-curried_aggregate = curried_aggregate_name (open_paren _ aggregate_list? _ close_paren) (open_paren _ filter _ close_paren)
-arbitrary_function = arbitrary_function_name (open_paren ( _ filter _ ) (_ comma _ param_expression)* close_paren)
+aggregate_name = "avg" / "count" / "max" / "min" / "sum" / "last" / "uniq"
+arbitrary_function = arbitrary_function_name (open_paren ( _ expression _ ) (_ comma _ expression)* close_paren)
+arbitrary_function_name = ~r"[a-zA-Z0-9_]+"
+curried_aggregate = curried_aggregate_name (open_paren _ aggregate_list? _ close_paren) (open_paren _ expression (_ comma _ expression)* _ close_paren)
```
evanh, line 60: I don't think this is right. A curried_aggregate should only have one parameter, which is a string. So `quantiles(0.5)(transaction.duration)` is a curried aggregate, but `quantiles(0.5)(transaction.duration, 10)` is not.

evanh: I suspect you will need a `curried_arbitrary_function` to support expressions like `topK(10)(sum(transaction.duration))`

snuba_sdk/timeseries.py
```diff
@@ -84,7 +84,7 @@ class Timeseries:
     """
 
     metric: Metric
-    aggregate: str
+    aggregate: str | None = None
```
evanh, line 87: I don't think this is a good idea. I actually disagree with the strategy you're taking here: I think `p90(transaction.duration)` should be `Formula(name="p90", parameters=tuple("transaction.duration")`. The reason is that I think a Timeseries should always have an aggregate.

evanh (changes requested):

snuba_sdk/mql/mql.py
enochtangg: That makes sense. I can add `curried_arbitrary_function`.

snuba_sdk/timeseries.py
enochtangg: Sure, the problem with that is we lose support for filtering and grouping by inside aggregate. 
For example: `p90(transaction.duration{foo:bar})`

In the suggested strategy, this would become `Formula(name="p90", parameters=tuple("transaction.duration{foo:bar}")`. My thought is that we resolve into a `Timeseries` (with an empty `aggregate` for now) we could plug the `quantiles` aggregate later in the metrics layer.

enochtangg:
From discussion offline:
* Use `""` to determine whether parameter is a string. Unquoted strings are treated as metric names.
* An aggregate (timeseries) must contain only one parameter and must be a metric. With this rule in place, aggregate names can be anything.
* If an arbitrary function param is a metric, it must also contain an aggregate. Else, the query is invalid and throw an error.

See `test_mql.py` for list of supported MQL strings.

snuba_sdk/mql/mql.py
```diff
@@ -53,21 +53,25 @@
...
-aggregate = aggregate_name (open_paren _ filter _ close_paren)
-curried_aggregate = curried_aggregate_name (open_paren _ aggregate_list? _ close_paren) (open_paren _ filter _ close_paren)
-arbitrary_function = arbitrary_function_name (open_paren ( _ filter _ ) (_ comma _ param_expression)* close_paren)
+function = (curried_aggregate / curried_arbitrary_function / aggregate / arbitrary_function) (group_by)?
+aggregate = aggregate_name (open_paren _ inner_filter _ close_paren)
+aggregate_name = ~r"[a-zA-Z0-9_]+"
+arbitrary_function = arbitrary_function_name (open_paren ( _ expression _ ) (_ comma _ expression)* close_paren)
+arbitrary_function_name = ~r"[a-zA-Z0-9_]+"
+curried_aggregate = curried_aggregate_name (open_paren _ aggregate_list? _ close_paren) (open_paren _ inner_filter _ close_paren)
+curried_aggregate_name = "quantiles" / "histogram"
```
evanh, line 62: Why is this still hardcoded? If we can, I would treat this the same way as regular aggregates.

tests/test_mql.py
```diff
@@ -1001,3 +1105,183 @@ def test_parse_mql_arbitrary_functions(
...
+                        ],
+                    ),
+                ],
+            )
+        ),
+        id="test curried arbitrary function with inner aggregate and terms",
+    ),
+    pytest.param(
+        "topK(10)(sum(transaction.duration{bar:baz}) / count(transaction.duration{foo:foz})) by transaction",
+        MetricsQuery(
```
evanh, line 1212: Could you add some of these more complex examples to the serializing tests as well? `test_metrics_mql_query.py`? Just to make sure we are properly serializing these things.

tests/test_mql.py
```diff
@@ -1001,3 +1105,183 @@ def test_parse_mql_arbitrary_functions(
 ) -> None:
     result = parse_mql(mql_string)
     assert result == metrics_query
+
+
+curried_arbitrary_function_tests = [
+    pytest.param(
+        "topK(10)(transaction.duration)",
```
evanh, line 1112: It seems like, based on the grammar, this test should fail, since this doesn't match a `curried_aggregate`. I think it does pass because you have some smart logic in the Visitor class, but it would be cleaner if it was the grammar that was making this work.

evanh (commented):
One test I don't see is `p90(transaction.duration)`. Is that capture somewhere already?

snuba_sdk/mql/mql.py
enochtangg: Are you referring to the `curried_aggregate`? We can treat them as `curried_arbitrary_function`.

tests/test_mql.py
enochtangg: It matches `curried_arbitrary_function`

tests/test_mql.py
enochtangg: yes, I can add this 👍

snuba_sdk/mql/mql.py
enochtangg: Removed `curried_aggregate`.

evanh (approved):

snuba_sdk/mql/mql.py
enochtangg: Updated to allow arbitrary curried names.

PR #164, opened 2024-01-19
fix(docs) Keep READMEs in sync

evanh:


evanh:
I don't actually remember why we have 2 READMEs now. Can probably remove one.

PR #165, opened 2024-01-25
feat(mql): Remove all metrics to SnQL translation code

evanh:
Now that we can fully use MQL in Snuba for all supported queries, we don't need
the SnQL translation at all.


volokluev (approved):
🥲 🥲 🥲 🥲  brings tears to my eyes. What a beautiful PR.

PR #168, opened 2024-01-31
feat(metrics): Allow uppercase metric names

iambriccardo:
This PR adds support of uppercase metric names in the snuba-sdk MQL grammar.

This change is linked to: https://github.com/getsentry/sentry/pull/64250


mitsuhiko (approved):
Relay emits these, so this was a divergence between what relay can produce and what we enable in querying.

PR #170, opened 2024-02-06
feat: Support MQL strings in MetricsQuery

evanh:
Allow providing a MQL string as the `query` parameter to a MetricsQuery. The
string will get parsed and validated when being serialized to ensure that it's
a valid MQL string.

Since an MQL string can't specify an entity, this requires that Snuba can
automatically infer the entity based on the passed in MRI. See https://github.com/getsentry/snuba/pull/5501

This also refactored a lot of tests: `parse_mql` was wrapping the parsed
value in a MetricsQuery object. Returning the Formula/Timeseries directly
meant gently refactoring a lot of tests.

### DIFF
--- a/snuba_sdk/metrics_visitors.py
+++ b/snuba_sdk/metrics_visitors.py
@@ -44,7 +44,7 @@ class TimeseriesVisitor(ABC, Generic[TVisited]):
     @abstractmethod
     def _visit_metric(
         self, metric: Metric
-    ) -> str | Mapping[str, TVisited | Mapping[str, TVisited]]:
+    ) -> str | Mapping[str, TVisited | Mapping[str, TVisited | None]]:
         raise NotImplementedError
 
     @abstractmethod
@@ -126,7 +126,9 @@ class TimeseriesMQLPrinter(TimeseriesVisitor[str]):
             "entity": metric_data["entity"],
         }
 
-    def _visit_metric(self, metric: Metric) -> Mapping[str, str | Mapping[str, str]]:
+    def _visit_metric(
+        self, metric: Metric
+    ) -> Mapping[str, str | Mapping[str, str | None]]:
         return self.metrics_visitor.visit(metric)
 
     def _visit_aggregate(
@@ -212,14 +214,17 @@ class MetricVisitor(ABC, Generic[TVisited]):
         raise NotImplementedError
 
 
-class MetricMQLPrinter(MetricVisitor[Mapping[str, Union[str, Mapping[str, str]]]]):
-    def visit(self, metric: Metric) -> Mapping[str, str | Mapping[str, str]]:
+class MetricMQLPrinter(
+    MetricVisitor[Mapping[str, Union[str, Mapping[str, Union[str, None]]]]]
+):
+    def visit(
+        self, metric: Metric
+    ) -> Mapping[str, str | Mapping[str, Union[str, None]]]:
         if metric.mri is None and metric.public_name is None:
             raise InvalidExpressionError(
                 "metric.mri or metric.public is required for serialization"
             )
-        if metric.entity is None:
-            raise InvalidExpressionError("metric.entity is required for serialization")
+
         if metric.mri:
             return {"metric_name": metric.mri, "entity": {metric.mri: metric.entity}}
         assert metric.public_name is not None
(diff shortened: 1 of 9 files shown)


enochtangg (commented):
If we support MQL strings in `MetricsQuery`, how does this work with `run_query()` in sentry? Specifically with resolving the metric_id.

evanh:
> If we support MQL strings in `MetricsQuery`, how does this work with `run_query()` in sentry? Specifically with resolving the metric_id.

The `run_query` function will decode the string into the SDK objects, and then continue with the normal flow from there.

snuba_sdk/metrics_visitors.py
```diff
@@ -212,14 +214,17 @@ def visit(self, metric: Metric) -> TVisited:
...
 
 
-class MetricMQLPrinter(MetricVisitor[Mapping[str, Union[str, Mapping[str, str]]]]):
-    def visit(self, metric: Metric) -> Mapping[str, str | Mapping[str, str]]:
+class MetricMQLPrinter(
+    MetricVisitor[Mapping[str, Union[str, Mapping[str, Union[str, None]]]]]
+):
+    def visit(
+        self, metric: Metric
+    ) -> Mapping[str, str | Mapping[str, Union[str, None]]]:
```
volokluev, line 222: If we could make this type more descriptive that would be cool. Right now it's complex and I have a hard time understanding what this function is actually returning

snuba_sdk/mql/mql.py
```diff
@@ -13,10 +13,14 @@
 from snuba_sdk.column import Column
 from snuba_sdk.conditions import And, BooleanCondition, BooleanOp, Condition, Op, Or
 from snuba_sdk.formula import ArithmeticOperator, Formula
-from snuba_sdk.metrics_query import MetricsQuery
-from snuba_sdk.query_visitors import InvalidQueryError
 from snuba_sdk.timeseries import Metric, Timeseries
 
+
+# In order to avoid circular imports, we need to redefine the exception here.
```
volokluev, line 19: you could just define an `exceptions` module as well and then you don't have to redefine

snuba_sdk/mql/mql.py
```diff
@@ -107,7 +111,7 @@
 }
 
 
-def parse_mql(mql: str) -> MetricsQuery:
+def parse_mql(mql: str) -> Timeseries | Formula:
```
volokluev, line 114: the docstring is now out of date. I'm also not sure how or where MetricsQuery is constructed now. if not here

snuba_sdk/mql_visitor.py
```diff
@@ -105,14 +102,17 @@ def _combine(
         }
 
     def _visit_query(
-        self, query: Timeseries | Formula | None
+        self, query: Timeseries | Formula | str | None
     ) -> Mapping[str, str | Mapping[str, str]]:
         if query is None:
             raise InvalidMetricsQueryError("MetricQuery.query must not be None")
```
volokluev, line 108: why is None a valid type in the type signature if it'll just throw an exception?

volokluev (approved):
Code is overall good but I have some questions about the changes

snuba_sdk/mql_visitor.py
evanh: Because we want to allow people to instantiate these objects without exceptions being thrown. This allows people to build these objects iteratively. This check happens when we actually serialize the object, since at that point we need a value here.

snuba_sdk/mql/mql.py
evanh: The MetricsQuery would be constructed by whoever calls this function. The reason I removed this is because wrapping the parsed object in a MetricsQuery was largely useless. The MetricsQuery has nothing populated besides the `query` field, and is usually discarded anyways.

For example: I use this function in the visitor when validating that the MQL string is correct. That would mean calling `parse_mql` inside the `MetricsQuery` visitor,  and then discarding the `MetricsQuery` anyways.

snuba_sdk/metrics_visitors.py
evanh: Honestly the typing of these functions needs to be reworked in general. These functions are tied to the base functions of the generic class, and that has been a problem in general with metrics, since we want to be able to return more complex types and not be locked into the generic type defined for the class.

volokluev (approved):

PR #177, opened 2024-02-20
fix(mql): Fix ambiguous grammar for filters

iambriccardo:
This PR fixes the grammar for filters, which misused quantifiers allowing filters without being surrounded by `{}`. The new grammar makes filter definitions unambiguous.

An example that would show the ambiguity of the grammar is as follows:
`rate(count(g:custom/zone.domains@none))`
which results in:
`Timeseries(metric=Metric(public_name='count', mri=None, id=None), aggregate='rate', aggregate_params=None, filters=[Condition(lhs=Column(name='g', entity=None, subscriptable=None, key=None), op=<Op.EQ: '='>, rhs='custom/zone.domains@none')], groupby=None)` which is clearly wrong.

The new grammar will output the correct result, that is, a `Formula` of `function_name=rate` containing a `Timeseries`.

### DIFF
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -37,7 +37,7 @@ term_op = "*" / "/"
 coefficient = number / quoted_string / filter
 
 number = ~r"[0-9]+" ("." ~r"[0-9]+")?
-filter = target open_brace? (_ filter_expr _ )? close_brace? (group_by)?
+filter = target (open_brace (_ filter_expr _ )? close_brace)? (group_by)?
 
 filter_expr = filter_term (_ or _ filter_term)*
 filter_term = filter_factor (_ joint_operator? _ filter_factor)*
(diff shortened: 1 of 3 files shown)


snuba_sdk/mql/mql.py
```diff
@@ -37,7 +37,7 @@ class InvalidMQLQueryError(Exception):
 coefficient = number / quoted_string / filter
 
 number = ~r"[0-9]+" ("." ~r"[0-9]+")?
-filter = target open_brace? (_ filter_expr _ )? close_brace? (group_by)?
+filter = target (open_brace (_ filter_expr _ )? close_brace)? (group_by)?
```
iambriccardo, line 40: This definition was wrong in the first place since it implies that filters can exist even without being surrounded by `{}` making the grammar ambiguous.

evanh (approved):
Thanks for this! Can you add an entry in CHANGES for the release?

iambriccardo:
> Thanks for this! Can you add an entry in CHANGES for the release?

Sure!

PR #179, opened 2024-03-07
feat(mql): Add support for unary operator

iambriccardo:
This PR adds support for unary operator `-` on numeric scalars and timeseries/formulas, which was requested by some users.

### DIFF
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -114,7 +114,7 @@ TERM_OPERATORS: Mapping[str, str] = {
 }
 
 UNARY_OPERATORS: Mapping[str, str] = {
-    "-": ArithmeticOperator.MINUS.value,
+    "-": "negate",
 }
 
 
@@ -199,7 +199,7 @@ class MQLVisitor(NodeVisitor):  # type: ignore
             elif isinstance(coefficient, Formula) or isinstance(
                 coefficient, Timeseries
             ):
-                return Formula(function_name=unary_op[0], parameters=[0, coefficient])
+                return Formula(function_name=unary_op[0], parameters=[coefficient])
             else:
                 raise InvalidMQLQueryError(
                     f"Unary expression not supported for type {type(coefficient)}"
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -194,12 +194,15 @@ class MQLVisitor(NodeVisitor):  # type: ignore
     ) -> Union[Formula, Timeseries, float, int, str]:
         unary_op, coefficient = children
         if unary_op:
-            # We want to transform -1 in 0 - 1.
             if isinstance(coefficient, float) or isinstance(coefficient, int):
                 return -coefficient
+            elif isinstance(coefficient, Formula) or isinstance(
+                coefficient, Timeseries
+            ):
+                return Formula(function_name=unary_op[0], parameters=[0, coefficient])
             else:
                 raise InvalidMQLQueryError(
-                    "Unary expression only applicable on a numeric scalar"
+                    f"Unary expression not supported for type {type(coefficient)}"
                 )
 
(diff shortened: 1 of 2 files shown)


snuba_sdk/mql/mql.py
```diff
@@ -168,21 +176,40 @@ def visit_term(
...
     def visit_term_op(self, node: Node, children: Sequence[Any]) -> Any:
         return TERM_OPERATORS[node.text]
 
+    def visit_unary(
+        self, node: Node, children: Sequence[Any]
+    ) -> Union[Formula, Timeseries, float, int, str]:
+        unary_op, coefficient = children
+        if unary_op:
+            if isinstance(coefficient, float) or isinstance(coefficient, int):
+                return -coefficient
```
iambriccardo, line 198: We can more efficiently negate directly the scalar instead of using a formula.

snuba_sdk/mql/mql.py
```diff
@@ -32,8 +32,12 @@ class InvalidMQLQueryError(Exception):
 expression = term (_ expr_op _ term)*
 expr_op = "+" / "-"
 
-term = coefficient (_ term_op _ coefficient)*
+term = unary (_ term_op _ unary)*
 term_op = "*" / "/"
+
+unary = unary_op? coefficient
+unary_op = "-"
```
evanh, line 39: I think the `number` rule should be changed to catch the negative sign. That will make negative scalars (`-1`) parse in a more logical way.

snuba_sdk/mql/mql.py
```diff
@@ -109,6 +113,10 @@ class InvalidMQLQueryError(Exception):
     "/": ArithmeticOperator.DIVIDE.value,
 }
 
+UNARY_OPERATORS: Mapping[str, str] = {
+    "-": ArithmeticOperator.MINUS.value,
```
evanh, line 117: Clickhouse uses the `negate` function for this, Clickhouse expects the `minus` function to have 2 arguments.

snuba_sdk/mql/mql.py
```diff
@@ -168,21 +176,40 @@ def visit_term(
...
+        self, node: Node, children: Sequence[Any]
+    ) -> Union[Formula, Timeseries, float, int, str]:
+        unary_op, coefficient = children
+        if unary_op:
+            if isinstance(coefficient, float) or isinstance(coefficient, int):
+                return -coefficient
+            elif isinstance(coefficient, Formula) or isinstance(
+                coefficient, Timeseries
+            ):
+                return Formula(function_name=unary_op[0], parameters=[0, coefficient])
```
evanh, line 202: Why does this have two parameters? Is that to support `minus()`? If so, we should use `negate()`.

evanh (commented):
Some changes necessary here I think.

Also a couple test cases I would add:

- `x - -y` Two minus signs next to each other
- `-sum(mri)` Should output a Formula instead of a Timeseries

snuba_sdk/mql/mql.py
iambriccardo: Oh I didn't know there was that function, will use that!

snuba_sdk/mql/mql.py
iambriccardo: In my implementation I support negation for numeric scalars and formulas/timeseries.

evanh: That's true, but if you added changed the definition of a `number` to support negatives, you wouldn't need to manually negate the numbers in the `unary` code.

Also it would just conceptually make more sense: the `number` rule defines a number, instead of needing a the `number` and the `unary` rule in order to get a correct scalar value.

iambriccardo: Decided to keep the grammar as is since adding negation multiple times in the grammar is less clean.

evanh (approved):

PR #180, opened 2024-03-12
fix(mql): Add support for negation in mql visitor

iambriccardo:
This PR implements the missing part of the unary implementation, that is, the serialization of `-` when generating mql from the ast.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changelog and versioning
 ==========================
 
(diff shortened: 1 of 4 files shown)


evanh (approved):
Please also update CHANGES with the new versions.

iambriccardo:
> Please also update CHANGES with the new versions.

Will do, thanks for reminding me! I always forget

PR #183, opened 2024-04-04
feat: sdk or optimizer

kylemumma:
This PR adds an optimizer pass to the Query and MetricsQuery during serialization. 
It will take a condition: `c=val1 or c=val2 or c=val3...` and convert it to `c in [val1, val2, val3, ...]`.
This change should fix --at least some of the queries causing-- the recursion bug in parsing.

## testing
* I wrote unit tests for the OrOptimizer to verify it optimizes ConditionGroups as expected. 
* I wrote more "end-to-end" tests of `serialize` that verify Query and MetricsQuery are optimized as expected when serialize is called.
* I wrote a test in sentry that sends an MQL query that caused the recursion bug previously. This test doesn't cause the error anymore. Note, this test is not committed to any repo it only exists locally in my own branch. https://github.com/getsentry/sentry/blob/560fe2f33afe0448b13ad5cab0c71d077309e0d5/tests/sentry/api/endpoints/test_organization_metrics_query.py#L110

Concerns:
* Do I need to run all of sentry CI test against the new sdk before i deploy?
* IDK what to do after merging this to SDK

### DIFF
--- a/snuba_sdk/metrics_query.py
+++ b/snuba_sdk/metrics_query.py
@@ -7,7 +7,7 @@ from typing import Any
 
 from snuba_sdk.expressions import Limit, Offset
 from snuba_sdk.formula import Formula
-from snuba_sdk.metrics_query_visitors import Validator
+from snuba_sdk.metrics_query_visitors import OrOptimizer, Validator
 from snuba_sdk.mql_visitor import MQLPrinter
 from snuba_sdk.query import BaseQuery
 from snuba_sdk.query_visitors import InvalidQueryError
@@ -92,8 +92,17 @@ class MetricsQuery(BaseQuery):
 
     def serialize(self) -> str | dict[str, Any]:
         self.validate()
+        self._optimize()
         result = MQL_PRINTER.visit(self)
         return result
+
+    def _optimize(self) -> None:
+        if isinstance(self.query, str):
+            raise ValueError(
+                "Expected a Formula or Timeseries query, got a str. Hint: call validate before optimizing?"
+            )
+        if self.query is not None:
+            self.query = OrOptimizer().optimize(self.query)
 
 
 MQL_PRINTER = MQLPrinter()
--- a/snuba_sdk/metrics_query.py
+++ b/snuba_sdk/metrics_query.py
@@ -98,8 +98,13 @@ class MetricsQuery(BaseQuery):
         return result
 
     def _optimize(self) -> None:
-        if isinstance(self.query, (Formula, Timeseries)):
-            self.query = OrOptimizer().optimize(self.query)
+        if (
+            isinstance(self.query, (Formula, Timeseries))
+            and self.query.filters is not None
+        ):
+            new_filters = OrOptimizer().optimize(self.query.filters)
+            if new_filters is not None:
+                self.query = replace(self.query, filters=new_filters)
 
 
 MQL_PRINTER = MQLPrinter()
--- a/snuba_sdk/metrics_query.py
+++ b/snuba_sdk/metrics_query.py
@@ -7,9 +7,10 @@ from typing import Any
 
 from snuba_sdk.expressions import Limit, Offset
 from snuba_sdk.formula import Formula
-from snuba_sdk.metrics_query_visitors import OrOptimizer, Validator
+from snuba_sdk.metrics_query_visitors import Validator
 from snuba_sdk.mql_visitor import MQLPrinter
 from snuba_sdk.query import BaseQuery
+from snuba_sdk.query_optimizers.or_optimizer import OrOptimizer
 from snuba_sdk.query_visitors import InvalidQueryError
 from snuba_sdk.timeseries import MetricsScope, Rollup, Timeseries
 
--- a/snuba_sdk/metrics_query.py
+++ b/snuba_sdk/metrics_query.py
@@ -98,11 +98,7 @@ class MetricsQuery(BaseQuery):
         return result
 
     def _optimize(self) -> None:
-        if isinstance(self.query, str):
-            raise ValueError(
-                "Expected a Formula or Timeseries query, got a str. Hint: call validate before optimizing?"
-            )
-        if self.query is not None:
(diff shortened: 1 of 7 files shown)


snuba_sdk/metrics_query.py
```diff
@@ -92,9 +93,18 @@ def print(self) -> str:
 
     def serialize(self) -> str | dict[str, Any]:
         self.validate()
+        self._optimize()
         result = MQL_PRINTER.visit(self)
         return result
 
+    def _optimize(self) -> None:
+        if isinstance(self.query, str):
```
evanh, line 101: Instead of throwing an exception here, I would simply skip this if the value is a str. It's possible for people to construct queries with raw MQL strings, but we parse them and re-encode them before sending to Snuba.

snuba_sdk/query_optimizers/or_optimizer.py
```diff
@@ -0,0 +1,68 @@
...
+            if res is not None:
+                optimized = True
+                new_filters.append(res)
+            else:
+                new_filters.append(cond)
+
+        if not optimized:
+            return query
+        elif isinstance(query, Timeseries):
+            return Timeseries(
```
evanh, line 29: This can be simplified a bit, the `dataclass` library provides the `replace` function:

```
replace(query, filters=new_filters)
```

This is equivalent to what you're doing here.

snuba_sdk/query_optimizers/or_optimizer.py
```diff
@@ -0,0 +1,68 @@
...
+                or (shared_lhs and curr.lhs != shared_lhs)
+            ):
+                # can't be optimized
+                return None
+            if not shared_lhs:
+                shared_lhs = curr.lhs
+            rhsides += [curr.rhs]
+        assert shared_lhs
+        assert rhsides
+        return Condition(shared_lhs, Op.IN, rhsides)  # type: ignore
```
evanh, line 68: Why is there a `# type: ignore` here?

snuba_sdk/query_optimizers/or_optimizer.py
```diff
@@ -0,0 +1,68 @@
...
+                not isinstance(curr, Condition)
+                or curr.op != Op.EQ
+                or (shared_lhs and curr.lhs != shared_lhs)
+            ):
+                # can't be optimized
+                return None
+            if not shared_lhs:
+                shared_lhs = curr.lhs
+            rhsides += [curr.rhs]
+        assert shared_lhs
```
evanh, line 66: Why these two asserts?

snuba_sdk/metrics_query.py
kylemumma: Ok I will make this change.

Another thought I had was if its a string you can parse first and then optimize, but this might make it slower and perhaps cause unexpected side-effects? Do you have any thoughts on this.

snuba_sdk/query_optimizers/or_optimizer.py
kylemumma: Thank Evan, I think I need to read up on dataclasses a bit

snuba_sdk/query_optimizers/or_optimizer.py
kylemumma: the new rhs for the `in` is a list of rhs, which works out to be of type
`Sequence[ScalarType | Expression]`
but rhs of a condition can only be of type
`Sequence[ScalarLiteralType | Expression]`

snuba_sdk/query_optimizers/or_optimizer.py
kylemumma: It was for mypy, I initialized them as None and empty outside of the loop, but they are guaranteed to be set inside the loop.

I can change it to initialize using first element, in order to remove the asserts. Do you think this would be better since it gets rid of the unnecessary asserts?

volokluev: getting rid of unnecessary asserts is best. You don't want a corner case to happen in production and break a bunch of queries

snuba_sdk/query_optimizers/or_optimizer.py
evanh: There's something funky with what you're doing here then, there should be a way to have this type correctly.

evanh (commented):
Are you planning to add the optimizer for SnQL in a separate PR?

tests/test_or_optimizer.py
```diff
@@ -0,0 +1,43 @@
+from snuba_sdk.column import Column
```
volokluev, line 1: Can you also add a test that makes sure that this optimization gets run for MQL queries and not SnQL queries? Basically something that asserts that the optimizer gets called when it's supposed to

volokluev (commented):
Great job. Just adda a more end-to-end test and we're golden

snuba_sdk/query_optimizers/or_optimizer.py
kylemumma: I don't think this is possible with the current type definitions. The best explanation I would be able to give for my reasoning over text would be what I wrote above.

RHS can be a sequence of literals, but it cannot be a sequence of sequence. Since a rhs can be a sequence, a list of rhs would be a sequence of sequence which is invalid.

Heres also a simple case to illustrate:
Screenshot 2024-04-09 at 4 02 45 PM

tests/test_or_optimizer.py
kylemumma: I added end-to-end tests for MQL and SnQL serialize

kylemumma:
## Changes
1. SnQL is now supported as well as MQL

2. The interface to OrOptimizer used to be
```
input: Formula | Timeseries
output: optimized Formula | Timeseries
```
now its
```
input: ConditionGroup
output: optimized ConditionGroup
```
I made this decision because when extending the optimizer to SnQL, it was the easiest path forward.

3. Added "end-to-end" tests of serialize, that verify SnQL and MQL queries optimize correctly when serialize is called.

4. Got rid of asserts

PR #184, opened 2024-04-08
feat(ddm): Add lenient MRI regex

iambriccardo:
This PR adds a more lenient MRI regex to relax the definition, as done in the backend: https://github.com/getsentry/sentry/pull/68175.

The definition of the unquoted MRI is slightly different than the one in the backend since we have to disambiguate certain strings which make the parser confused. The divergence of MRI regexes might cause problems in case some ill-defined metrics are accepted by Relay, the backend, and not by MQL. This however should not happen, but it is a possibility.

In the case of a quoted MRI, an even more lenient version is permitted, since we use backticks to disambiguate and we can therefore afford to put more accepted characters.

### DIFF
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -77,7 +77,8 @@ group_by_name_tuple = open_paren _ group_by_name (_ comma _ group_by_name)* _ cl
 
 inner_filter = metric (open_brace (_ filter_expr _)? close_brace)? (group_by)?
 metric = quoted_mri / unquoted_mri / quoted_public_name / unquoted_public_name
-quoted_mri = backtick unquoted_mri backtick
+quoted_mri = backtick lenient_mri backtick
+lenient_mri = ~r'[^:`]+:[^/`]+/[^@,`]+@[^`]+'
 unquoted_mri = ~r'[^:\(\){}\[\]"`,]+:[^/\(\){}\[\]"`,]+/[^@\(\){}\[\]"`,]+@[^\(\){}\[\]"`,]+'
 quoted_public_name = backtick unquoted_public_name backtick
 unquoted_public_name = ~r"([a-z_]+(?:\.[a-z_]+)*)"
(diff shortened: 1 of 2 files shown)


snuba_sdk/mql/mql.py
```diff
@@ -83,16 +78,16 @@ class InvalidMQLQueryError(Exception):
...
 unquoted_public_name = ~r"([a-z_]+(?:\.[a-z_]+)*)"
 
 open_paren = "("
 close_paren = ")"
 open_square_bracket = "["
 close_square_bracket = "]"
-open_brace = "{{"
-close_brace = "}}"
+open_brace = "{"
+close_brace = "}"
```
iambriccardo, line 90: Remove f-string, so this escaping is no more needed.

snuba_sdk/mql/mql.py
```diff
@@ -83,16 +78,16 @@ class InvalidMQLQueryError(Exception):
 inner_filter = metric (open_brace (_ filter_expr _)? close_brace)? (group_by)?
 metric = quoted_mri / unquoted_mri / quoted_public_name / unquoted_public_name
 quoted_mri = backtick unquoted_mri backtick
-unquoted_mri = ~r"{METRIC_TYPE_REGEX}:{METRIC_NAMESPACE_REGEX}/{METRIC_NAME_REGEX}@{METRIC_UNIT_REGEX}"
+unquoted_mri = ~r'[^:\(\){}\[\]"`,]+:[^/\(\){}\[\]"`,]+/[^@\(\){}\[\]"`,]+@[^\(\){}\[\]"`,]+'
```
iambriccardo, line 81: I have excluded all special chars of the grammar like `(){}[]"`,` to make sure the parser doesn't end up in an ambiguous state.

evanh (commented):
Can you add a test for this? Something that didn't parse before but should now?

Also, can you confirm what part of the system does or should guarantee that the metric type is valid?

iambriccardo:
> Can you add a test for this? Something that didn't parse before but should now?
> 
> Also, can you confirm what part of the system does or should guarantee that the metric type is valid?

Sure, will do.

Relay will guarantee this.

evanh (approved):

PR #186, opened 2024-05-07
feat(dev-exp): Add support for storage direct queries

enochtangg:
This PR is responsible adding support for storage direct queries in snuba. Now that entities are optional, queries can be made directly onto the snuba storage. As a result, we need the ability to write such queries. This PR adds a `Storage` class that can be references by the `Query`. The appropriate validators and tests were also added. 

Depends on: 
* https://github.com/getsentry/snuba/pull/5870


evanh (approved):

snuba_sdk/storage.py
```diff
@@ -0,0 +1,37 @@
...
+            elif self.sample <= 0.0:
+                raise InvalidStorageError("samples must be greater than 0.0")
+
+        if self.data_model is not None:
+            if not isinstance(self.data_model, DataModel):
+                raise InvalidStorageError("data_model must be an instance of DataModel")
+
+    def __repr__(self) -> str:
+        sample = f", sample={self.sample}" if self.sample is not None else ""
+        return f"Storage('{self.name}'{sample})"
```
volokluev, line 37: In the grammar it's written as `STORAGE` 
```suggestion
        return f"STORAGE('{self.name}'{sample})"
```

volokluev (approved):

PR #187, opened 2024-05-16
feat(metrics): add suffix wildcards & corresponding tests to filters in MQL grammar

shellmayr:
Contributes to https://github.com/getsentry/sentry/issues/69852

### DIFF
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -285,7 +285,7 @@ class MQLVisitor(NodeVisitor):  # type: ignore
         else:
             condition_op, lhs, _, _, _, rhs = factor
             op = Op.EQ
-            if not condition_op and isinstance(rhs, list) and node.text[-1]=='*':
+            if not condition_op and isinstance(rhs, list) and node.text[-1] == '*':
                 op = Op.LIKE
                 rhs = f"{rhs[0]}%"
 
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -299,7 +299,6 @@ class MQLVisitor(NodeVisitor):  # type: ignore
                 not condition_op
                 and contains_wildcard
                 and isinstance(rhs, str)
-                and contains_wildcard
             ):
                 op = Op.LIKE
 
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -46,9 +46,9 @@ nested_expr = open_paren _ filter_expr _ close_paren
 condition_op = "!"
 
 joint_operator = comma / and
- 
+
 tag_key = ~r"[a-zA-Z0-9_.]+"
-tag_value = quoted_suffix_wildcard_tag_value / suffix_wildcard_tag_value / quoted_string_filter / unquoted_string_filter / string_tuple / variable 
+tag_value = quoted_suffix_wildcard_tag_value / suffix_wildcard_tag_value / quoted_string_filter / unquoted_string_filter / string_tuple / variable
 suffix_wildcard_tag_value = unquoted_string wildcard
 quoted_suffix_wildcard_tag_value = quote unquoted_string wildcard quote
 
--- a/snuba_sdk/mql/mql.py
+++ b/snuba_sdk/mql/mql.py
@@ -353,14 +353,14 @@ class MQLVisitor(NodeVisitor):  # type: ignore
         self, node: Node, children: Sequence[Any]
     ) -> tuple[bool, str]:
         _, text_before_wildcard, _, _ = children
-        rhs = f"{text_before_wildcard}%"
+        rhs = f"{text_before_wildcard}*"
         return True, rhs
 
     def visit_suffix_wildcard_tag_value(
         self, node: Node, children: Sequence[Any]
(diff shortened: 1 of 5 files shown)


evanh (commented):
Why did you add new rules to the grammar for this? Could you also check to see if the node text ends with "*"?

iambriccardo:
> Why did you add new rules to the grammar for this? Could you also check to see if the node text ends with "*"?

We had a chat with @shellmayr about this and we agreed that we would like to encode wildcards in the grammar explicitly so that we have the same errors as any parsing error in case a string is not correct. In addition, we wanted to avoid any parsing logic inside the visitor since it would make it very easy to misimplement it.

snuba_sdk/mql/mql.py
```diff
@@ -276,20 +284,32 @@ def visit_filter_factor(
...
+            # condition_op, (variable or tag_key), space, colon, space, tag_value
+            condition_op, lhs, _, _, _, filter_factor_value = factor
+            contains_wildcard = filter_factor_value.contains_wildcard
+            rhs = filter_factor_value.value
             op = Op.EQ
-            if not condition_op and isinstance(rhs, list):
+            if not condition_op and contains_wildcard and isinstance(rhs, str):
+                if contains_wildcard:
+                    op = Op.LIKE
+
```
iambriccardo, line 301: nit: you could simplify it.
```suggestion
            if not condition_op and contains_wildcard and isinstance(rhs, str):
            	op = Op.LIKE
```

snuba_sdk/mql/mql.py
```diff
@@ -546,3 +593,9 @@ def visit_identifier(self, node: Node, children: Sequence[Any]) -> str:
     def generic_visit(self, node: Node, children: Sequence[Any]) -> Any:
         """The generic visit method."""
         return children
+
+
+@dataclass
+class FilterFactorValue(object):
```
iambriccardo, line 599: nit: no need to call it `*Value`, since it contains both value and metadata.
```suggestion
class FilterFactor(object):
```

iambriccardo (approved):
Great work Simon, left some minor nits!

snuba_sdk/mql/mql.py
shellmayr: Ah yeah that is in there twice, thank you for catching this!

snuba_sdk/mql/mql.py
shellmayr: Changed

shellmayr:
> Why did you add new rules to the grammar for this? Could you also check to see if the node text ends with "*"?

Riccardo's point is one part of it, but with every implicit extension the logic gets more and more opaque and hard to understand. I'd much prefer this being part of the grammar than encoded in a series of ifs and elses in the filter factor visitor.  Is there any reason we would NOT want this to be part of the grammar?

snuba_sdk/mql/mql.py
```diff
@@ -47,12 +49,17 @@ class InvalidMQLQueryError(Exception):
...
 tag_key = ~r"[a-zA-Z0-9_.]+"
-tag_value = quoted_string / unquoted_string / string_tuple / variable
+tag_value = quoted_suffix_wildcard_tag_value / suffix_wildcard_tag_value / quoted_string_filter / unquoted_string_filter / string_tuple / variable
+suffix_wildcard_tag_value = unquoted_string wildcard
+quoted_suffix_wildcard_tag_value = quote unquoted_string wildcard quote
 
-quoted_string = ~r'"([^"\\]*(?:\\.[^"\\]*)*)"'
-unquoted_string = ~r'[^,\[\]\"}{\(\)\s]+'
+quoted_string_filter = ~r'"([^"\\]*(?:\\.[^"\\]*)*)"'
+unquoted_string_filter = ~r'[^,\[\]\"}{\(\)\s\*]+'
```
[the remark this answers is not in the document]
shellmayr: Yeah I see what you mean, but I don't think there is a way to actually do this without changing the rest of the grammar that needs those quoted_string filters I think; aliasing itself does not work unfortunately, as the execution immediately jumps to the qutoed_string visitor instead.

shellmayr:
> > Why did you add new rules to the grammar for this? Could you also check to see if the node text ends with "*"?
> 
> Riccardo's point is one part of it, but with every implicit extension the logic gets more and more opaque and hard to understand. I'd much prefer this being part of the grammar than encoded in a series of ifs and elses in the filter factor visitor. Is there any reason we would NOT want this to be part of the grammar?

@evanh what do you think?

evanh:
> Riccardo's point is one part of it, but with every implicit extension the logic gets more and more opaque and hard to understand. I'd much prefer this being part of the grammar than encoded in a series of ifs and elses in the filter factor visitor. Is there any reason we would NOT want this to be part of the grammar?

This makes sense to me. The reason I mentioned it was that this will require changing the Snuba parser as well to handle the new rules.

evanh (approved):

shellmayr:
Yeah, I realize that that is some extra work but I think it is worth it to not let the parser logic degrade. FWIW there is already a prepared PR for snuba to integrate this logic (failing tests in CI because the current snuba-sdk does not yet have the updated grammar): https://github.com/getsentry/snuba/pull/5972/files

PR #189, opened 2024-07-22
bug(mql): Update rollup totals validation

enochtangg:
A Rollup object should be invalid if `interval=None` and `totals=None | False`. Since in MQL, a False or None rollup query mean the same thing.


evanh (commented):
Have you tested this with the snuba/sentry repo? Something tells me this might break some existing product logic.

enochtangg:
I thought this stricter validation would necessary to fix certain queries, but turns out this isn't actually required to fix formula totals queries with no groupbys: https://github.com/getsentry/snuba/pull/6136

PR #190, opened 2024-07-23
ci: dependency review action

mdtro:
Add dependency review action. Any new dependencies added that have a known high or critical vulnerability will fail CI.


mdtro:
Going to deploy this via global workflows.

PR #192, opened 2024-08-06
feat: new BaseQuery, DeleteQuery which will be used to make delete queries

kylemumma:
This PR adds a new subclass of `BaseQuery`, called `DeleteQuery`. This will be used inside a `Request` object for sdk users to create delete queries to send to snuba. Like this:

```
# create a snuba-sdk request object, query type DeleteQuery
request = Request(
    dataset = "search_issues",
    app_id = "myappid",
    tenant_ids = {"referrer": "my_referrer", "organization_id": 1234}
    query = DeleteQuery(
		    storage="search_issues"
		    columnConditions={
			    "project_id": [1],
			    "occurrence_id": UUID
		    }
    )
)

# use a function like raw_snql_query to send the request object to snuba
raw_snql_query(request, ...)
```

Next Steps: 
* change `raw_snql_query` to know what to do when the query type of a request is DeleteQuery
* update snuba to the new sdk

### DIFF
--- a/snuba_sdk/delete_query.py
+++ b/snuba_sdk/delete_query.py
@@ -1,0 +1,46 @@
+import json
+from dataclasses import dataclass
+from typing import Any, Dict, List, Union
+
+from snuba_sdk.query import BaseQuery
+
+
+@dataclass(frozen=True)
+class DeleteQuery(BaseQuery):
+    """
+    This represents a snuba delete query.
+    Inputs:
+        storage - the storage to delete from
+        columnConditions - a mapping from column-name to a list of column values
+            that defines the delete conditions. ex:
+            {
+                "id": [1, 2, 3]
+                "status": ["failed"]
+            }
+            represents
+            DELETE FROM ... WHERE id in (1,2,3) AND status='failed'
+    Deletes all rows in the given storage, that satisfy the conditions
+    defined in 'columnConditions'.
+    """
+
+    storage_name: str
+    column_conditions: Dict[str, List[Any]]
+
+    def validate(self) -> None:
+        """
+        i dont think we need to do any input validation at the sdk
+        level bc inputs will be validated at the endpoint and the proper
+        response code returned.
+        """
+        return
+
+    def serialize(self) -> Union[str, Dict[str, Any]]:
+        # the body of the request
+        self.validate()
+        return json.dumps({"columns": self.column_conditions})
+
+    def print(self) -> str:
+        return repr(self)
+
+    def __repr__(self) -> str:
+        return f"DeleteQuery(storage_name={repr(self.storage_name)}, columnsConditions={repr(self.column_conditions)})"
--- a/snuba_sdk/delete_query.py
+++ b/snuba_sdk/delete_query.py
@@ -31,12 +31,14 @@ class DeleteQuery(BaseQuery):
     column_conditions: Dict[str, List[Union[str, int]]]
 
     def validate(self) -> None:
-        if "project_id" not in self.column_conditions:
-            raise InvalidDeleteQueryError(
-                "missing required column condition on 'project_id'"
-            )
-        elif len(self.column_conditions["project_id"]) == 0:
-            raise InvalidDeleteQueryError("column condition on 'project_id' is empty")
+        if self.column_conditions == {}:
+            raise InvalidDeleteQueryError("column conditions cannot be empty")
+
+        for col, values in self.column_conditions.items():
+            if len(values) == 0:
+                raise InvalidDeleteQueryError(
+                    f"column condition '{col}' cannot be empty"
+                )
 
     def serialize(self) -> Union[str, Dict[str, Any]]:
         # the body of the request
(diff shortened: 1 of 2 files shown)


snuba_sdk/delete_query.py
```diff
@@ -0,0 +1,46 @@
...
+            represents
+            DELETE FROM ... WHERE id in (1,2,3) AND status='failed'
+    Deletes all rows in the given storage, that satisfy the conditions
+    defined in 'columnConditions'.
+    """
+
+    storage_name: str
+    column_conditions: Dict[str, List[Any]]
+
+    def validate(self) -> None:
```
kylemumma, line 29: I made the decision that input validation should not be done by the sdk, this is why I made that decision:
1. Input validation needs to be done (and is) by the actual snuba endpoint `/deletes`. By adding validation here as well we not only are writing duplicate code, but we are also creating 2 separate things that we need to keep updated now as opposed to 1.
2. The SDK already enforces proper types (with python typing and mypy) as long as these are respected then the client will be able to successfully deliver the request to the endpoint and recieve the proper response.

tests/test_delete_query.py
```diff
@@ -0,0 +1,31 @@
...
+            column_conditions={"project_id": [1], "occurrence_id": ["1234"]},
+        ),
+        '{"columns":{"project_id": [1], "occurrence_id": ["1234"]}}',
+    ),
+    pytest.param(
+        DeleteQuery(
+            storage_name="non-real-storage",
+            column_conditions={},
+        ),
+        '{"columns":{}}',
```
kylemumma, line 20: heres a practical example of my decision not to do validation at the sdk level.

the client doesnt include `project_id`, since theres no validation in the sdk it fine here, and the client will get 400 from the endpoint for missing project_id. If the requirements for the endpoint ever change its just 1 change at the endpoint.

The alternative where the sdk does validation would require project_id here, but if that ever changes it will require a code change at the endpoint as well as here. Which is why i think its better for validation to only happen at the endpoint.

snuba_sdk/delete_query.py
```diff
@@ -0,0 +1,46 @@
...
+        i dont think we need to do any input validation at the sdk
+        level bc inputs will be validated at the endpoint and the proper
+        response code returned.
+        """
+        return
+
+    def serialize(self) -> Union[str, Dict[str, Any]]:
+        # the body of the request
+        self.validate()
+        return json.dumps({"columns": self.column_conditions})
```
kylemumma, line 40: storage is in the url not the body, which is why its not in the serialize. it will be grabbed off the object later by `raw_snql_query` to put into the url.

snuba_sdk/delete_query.py
```diff
@@ -0,0 +1,46 @@
...
+                "status": ["failed"]
+            }
+            represents
+            DELETE FROM ... WHERE id in (1,2,3) AND status='failed'
+    Deletes all rows in the given storage, that satisfy the conditions
+    defined in 'columnConditions'.
+    """
+
+    storage_name: str
+    column_conditions: Dict[str, List[Any]]
```
MeredithAnya, line 27: I don't think we should use `Any` here, I think we should explicitly write out the options we allow so far. For now it's just `str` and `int` right?

snuba_sdk/delete_query.py
MeredithAnya: I'm not sure if I agree that we shouldn't do validation here. I don't think it makes sense to send a query to snuba with something like `{"project_id": []}` for the columns, or to have columns be empty (`{}`) like you denote in the tests. 

It's true that we do validation in the endpoint as well, but it would be better to prevent making the request to snuba at all if we already know it's invalid. As for needing to keep these both updated, I would think that the two cases I raised above are not likely to change even if we add more validation on the endpoint.

snuba_sdk/delete_query.py
kylemumma: yeah as far as I know just str and int, done

snuba_sdk/delete_query.py
kylemumma: done

kylemumma:
I added validation to the SDK object and I restricted the type of column condition values to str and int. LMK if theres any additional validation you think should be added.

snuba_sdk/delete_query.py
```diff
@@ -0,0 +1,50 @@
...
+            DELETE FROM ... WHERE id in (1,2,3) AND status='failed'
+    Deletes all rows in the given storage, that satisfy the conditions
+    defined in 'columnConditions'.
+    """
+
+    storage_name: str
+    column_conditions: Dict[str, List[Union[str, int]]]
+
+    def validate(self) -> None:
+        if "project_id" not in self.column_conditions:
```
MeredithAnya, line 34: I was actually thinking of a validation that was more general, something like the following

```python
if self.column_conditions == {}:
    raise InvalidDeleteQueryError("column conditions cannot be empty")

for col, values in self.column_conditions.items():
    if len(values) == 0:
        raise InvalidDeleteQueryError(f"column condition {col} cannot be empty")
```

thoughts? that way we wouldn't have to update this code for every column we add

MeredithAnya (approved):

PR #194, opened 2024-08-12
fix: DeleteQuery, send query as a dict

kylemumma:
The SDK was used to connect sentry to a delete endpoint in snuba. We ran into an issue however where the SDK was serializing the body of the request into a str when our data is actually a dict. This change makes it send the object as a dict now instead of a str.

Additionally we want to not send the dataset since its not required for this endpoint.

### DIFF
--- a/snuba_sdk/request.py
+++ b/snuba_sdk/request.py
@@ -84,6 +84,10 @@ class Request:
             mql_context = serialized_mql["mql_context"]
             query = str(serialized_mql["mql"])
         elif isinstance(self.query, DeleteQuery):
+            """
+            for a DeleteQuery, the query is not a snql/mql string,
+            it is a dict
+            """
             return {
                 **flags,
                 "query": self.query.serialize(),
--- a/snuba_sdk/request.py
+++ b/snuba_sdk/request.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 import re
 from dataclasses import asdict, dataclass, field, fields
-from typing import Mapping
+from typing import Any, Dict, Mapping, Union
 
 from snuba_sdk.metrics_query import MetricsQuery
 from snuba_sdk.query import BaseQuery
@@ -81,9 +81,9 @@ class Request:
             serialized_mql = self.query.serialize()
             assert isinstance(serialized_mql, dict)  # mypy
             mql_context = serialized_mql["mql_context"]
-            query = str(serialized_mql["mql"])
+            query: Union[str, Dict[str, Any]] = str(serialized_mql["mql"])
         else:
-            query = str(self.query.serialize())
+            query = self.query.serialize()
 
         ret: dict[str, object] = {
             **flags,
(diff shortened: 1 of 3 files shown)


snuba_sdk/request.py
```diff
@@ -81,9 +81,9 @@ def to_dict(self) -> dict[str, object]:
             serialized_mql = self.query.serialize()
             assert isinstance(serialized_mql, dict)  # mypy
             mql_context = serialized_mql["mql_context"]
-            query = str(serialized_mql["mql"])
+            query: Union[str, Dict[str, Any]] = str(serialized_mql["mql"])
         else:
-            query = str(self.query.serialize())
```
kylemumma, line 86: I verified that this change doesn't have any logical consequences. It was only there before to appease mypy.

xurui-c (approved):

snuba_sdk/request.py
onkar: With this change `ret` on line 88 can have a key `"query"` with type `str` if it's a metrics query. The same key will have a type `dict` if it is not a metric query. Is this an intentional outcome? I wonder if it'd create surprises in serialization logic. Also, I would appreciate a bit more context on what motivated this change in the PR description. Thanks!

snuba_sdk/request.py
```diff
@@ -81,9 +81,9 @@ def to_dict(self) -> dict[str, object]:
             serialized_mql = self.query.serialize()
             assert isinstance(serialized_mql, dict)  # mypy
             mql_context = serialized_mql["mql_context"]
-            query = str(serialized_mql["mql"])
+            query: Union[str, Dict[str, Any]] = str(serialized_mql["mql"])
```
MeredithAnya, line 84: let's put this under `isinstance(self.query, DeleteQuery)` since the metrics query would actually never be a dict. and then we can add a comment about why the query is different for deletes than the other queries

MeredithAnya:
@kylemumma as per our conversation just wanted to note here that we actually don't need the dataset for the snuba delete api, so one option is to pop it from the `to_dict` result. The other option is to update the schema definition in snuba to allow for additional properties. will leave that up to you

snuba_sdk/request.py
```diff
@@ -82,6 +83,14 @@ def to_dict(self) -> dict[str, object]:
...
             mql_context = serialized_mql["mql_context"]
             query = str(serialized_mql["mql"])
+        elif isinstance(self.query, DeleteQuery):
+            return {
+                **flags,
+                "query": self.query.serialize(),
+                "app_id": self.app_id,
+                "tenant_ids": self.tenant_ids,
+                "parent_api": self.parent_api,
+            }
```
kylemumma, line 93: obviously this is a duct tape solution. Its what we will use for now since we kind of have to use the Request object right now.

A longer term solution would likely be to refactor this request object. Maybe considered down the line.

snuba_sdk/request.py
kylemumma: Hi Onkar thanks for taking the time to review this. Sorry for the bad description I wasnt expecting people without context to look at it. I added more context to the description so hopefully it makes more sense.

The PR has been changed slightly so now the type mismatch is not a thing anymore

MeredithAnya (approved):

onkar (approved):
Looks good, thanks for addressing the comments!

PR #195, opened 2024-08-26
feat(mql): Add extrapolate flag to MQL context

enochtangg:
Now that we are storing the weighted metrics, we want a mechanism to query these extrapolated metrics in MQL. This PR is responsible for adding an `extrapolate` flag to the MetricsQuery object attached in the mql context. By setting this flag to `True|False`, snuba will determine whether or not to query and return the weighted columns (extrapolated metrics). For example, the follow metrics query object:
```
MetricsQuery(
            query=Timeseries(
                metric=Metric(
                    mri="d:transactions/duration@millisecond",
                ),
                aggregate="max",
                aggregate_params=None,
                filters=[Condition(Column("bar"), Op.EQ, "baz")],
                groupby=[Column("transaction")],
            ),
            start=NOW,
            end=NOW + timedelta(days=14),
            rollup=Rollup(interval=3600, totals=None, granularity=3600),
            scope=MetricsScope(
                org_ids=[1], project_ids=[11], use_case_id="transactions"
            ),
            limit=Limit(100),
            offset=Offset(5),
            extrapolate=Extrapolate(True),
            indexer_mappings={},
        )
```
will now get serialized as:
```
{
            "mql": 'max(d:transactions/duration@millisecond){bar:"baz"} by (transaction)',
            "mql_context": {
                "start": "2023-01-02T03:04:05+00:00",
                "end": "2023-01-16T03:04:05+00:00",
                "rollup": {
                    "orderby": None,
                    "granularity": 3600,
                    "interval": 3600,
                    "with_totals": None,
                },
                "scope": {
                    "org_ids": [1],
                    "project_ids": [11],
                    "use_case_id": "transactions",
                },
                "limit": 100,
                "offset": 5,
                "extrapolate": True,
                "indexer_mappings": {},
            },
        }
```

Merging this PR and bumping the snuba-sdk version on snuba does not affect existing MQL queries. This is because the snuba's MQL API schema accepts a pretty generic type (see [here](https://github.com/getsentry/snuba/blob/84bf293d7e6c2c636718fbcbcfc0938dcfc33fb9/snuba/query/schema.py#L28)). As a result, this flag will be ignored for now until we make the parser changes in snuba to resolve the columns appropriately.


john-z-yang (approved):

tests/test_metrics_mql_query.py
```diff
@@ -57,6 +57,7 @@
                 },
                 "limit": None,
                 "offset": None,
+                "extrapolate": None,
```
john-z-yang, line 60: Some quick clarifying questions:
Do we have to pass in the `extrapolate` field? I think maybe we should make it so that the passing in the `extrapolate` field is optional, but if it is passed in, it needs to be either true or false

enochtangg: Yes, that is correct, the `extrapolate` field is not mandatory. There are tests in the `test_metrics_mql_query.py` file that do not pass `"extrapolate"` and still get serialize appropriately.

